PR #18, opened 2026-07-18
fix: update MapReduce estimate arity + TileCompleted pattern for current core (libviprs#327)

spdrman:
Threads the new `channel_bytes` argument into `compute_inflight_strips`, and updates the flamegraph `EngineEvent::TileCompleted` pattern for the new `worker_id`/`timestamp` fields, so bench builds against core main again. Also clears three incidental lints (redundant field, needless mut, dead fn). Ref libviprs/libviprs#327.


spdrman:
Closing: based on a stale local checkout. Current libviprs-bench `main` (21fa56a) already builds against core main. No change needed.

Issue #24, opened 2026-07-20
Dockerfile: build + record a modern pinned libvips


spdrman:
The container path benches against Debian's frozen libvips 8.14.1 with mismatched 8.18 bindings. Build libvips from a pinned recent source tarball (8.16/8.17.x) in the Dockerfile, align the libvips-rs bindings, and make LIBVIPS_VERSION a recorded axis. (4-expert review, major.)

PR #39, opened 2026-07-20
Version-matrix runner: --versions driver for the release-history axis (#19)

### LINKED ISSUE #19: Version-matrix runner: --versions <tags> driver for the release-history axis
The fairness rework did the cross_version half (schema migrate-on-load, environment-fingerprint grouping, CI-overlap regression gating). STILL NEEDED: a --versions <tag,tag,HEAD> driver (bin or run-bench.sh mode) that per libviprs version checks out ../libviprs into a temp git worktree at that tag, rebuilds the harness with the identical pinned profile, runs the identical suite, and appends one tagged+fingerprinted snapshot each — producing the release-history axis in ONE invocation instead of manual accretion. (4-expert review, high priority — this is the core 'compare different libviprs versions' feature.)

### LINKED ISSUE #26: version-matrix: per-tag worktree checkout + rebuild driver
Sub-issue of #19. Given a libviprs git tag/SHA, check out ../libviprs into a temp git worktree at that ref and rebuild the bench harness against it with the identical pinned profile. Deterministic + cleaned up after. Tests: assert the driver builds a known tag and reports its resolved SHA.

### LINKED ISSUE #27: version-matrix: --versions driver runs the suite + appends one tagged, fingerprinted snapshot per version
Sub-issue of #19. A `--versions <tag,tag,HEAD>` mode (bin or run-bench.sh) that, per version, runs the identical suite via the per-tag build (sub-issue), and appends ONE tagged + environment-fingerprinted BenchmarkSnapshot each — producing the release-history axis in one invocation. Tests: two-version dry run appends two distinct tagged snapshots that load_history reads back.

spdrman:
Resolves **#19** and sub-issues **#26** (per-tag worktree checkout + rebuild driver) and **#27** (--versions runner + tagged/fingerprinted append + version ordering).

New `version_matrix` module + bin + run-bench.sh `versions` mode: checks out ../libviprs per tag into a temp worktree, rebuilds the harness against it (identical pinned profile), runs the identical suite, and appends one tagged + environment-fingerprinted snapshot each — a build failure at an old tag is skip+warn, never aborting the matrix. Snapshot identity is **verified** against the built artifact (build.rs-stamped `report --print-core`), `save_history` is atomic, and `cross_version` now keys by version@short_sha and orders by (semver, timestamp) instead of lexicographically.

TDD: commit 1 failing tests (red: E0432/E0425); commit 2 implementation. 4-expert review applied (consolidated below). Rebased onto #24. Local feature-matrix gate green (CI skipped via [skip ci]).

Closes #19
Closes #26
Closes #27


spdrman:
## Consolidated review — Version-matrix runner (#19)

**Head:** `83437c8` (`bench/19-version-matrix`) — single commit `refactor: apply 4-expert review (#19) [skip ci]`.
**Verify:** `cargo build` ✅ · `cargo fmt --check` ✅ · `cargo clippy --all-targets -D warnings` ✅ (rc=0; the one printed "warning" is the pre-existing benign pdfium `[patch]`-unused note) · `cargo clippy --features polars --bin cross_version -D warnings` ✅ · `cargo test` ✅ (15 lib-unit + 7 integration, +2 ignored, 0 failures) · **real per-tag rebuild** ignored test ✅ (36s: checks out HEAD, compiles `libviprs 0.4.0` through the `paths` override, self-report-verifies, appends) · smokes ✅ (`--help`→0, missing `--versions`→2, `report --print-core`→`0.4.0\ta055719`, corrupt-history pre-flight→2 in 0.14s with no wasted rebuild).

### Overall assessment
Strong feature with solid TDD scaffolding. The four experts converged on two themes: **provenance integrity** (the snapshot's recorded identity was asserted from a side channel, not from the artifact that was actually benchmarked — provably wrong in `Reuse` mode) and **crash-safety** (non-atomic/`unwrap()`-ing history writes; worktree/temp-dir leaks on early-return or panic). Both are now closed.

| Axis | Before → after |
|---|---|
| Architecture | Pure identity/ordering logic extracted to a dependency-free `version_id` module; `cross_version` no longer compile-coupled to the git/cargo runner. |
| Implementation | Measured binary is now the identity source of truth (`--print-core`, verified post-build); suite parameters are shared `const`s, not hand-copied literals. |
| Safety | `save_history` atomic (temp+rename) + fallible; worktree guard armed before any fallible step; owned build tree RAII-cleaned; write faults degrade to a per-version skip, never abort the sweep. |
| Performance | Hot path unchanged; the deliberate shared target dir is kept, now guarded by a per-tag self-report re-verification that catches a stale-binary mislink. |
| Maintainability | Typed skip outcomes, distinct exit codes (1/3), deduped manifest scanner, per-version heartbeat + wall-clock. |
| Production-readiness | Fit for purpose (host-measured, single-writer release-history axis), now explicitly documented as such. |

### Mandatory (all applied)
- **Artifact-sourced identity** *(Architect#1/#7, RustSys#5, Safety#4, API#5 — High).* Added hidden `report --print-core` emitting the `build.rs`-stamped `core_version()\tcore_git_sha()`. `build_harness` (Rebuild) now asserts the freshly-built binary's self-report matches the resolved ref → `MatrixError::Build` on mismatch, making the `BENCH_CORE_DIR` plumbing load-bearing and catching a stale stamp/binary in the shared target dir. `Reuse`→**`ReuseUnchecked`** with a loud invariant doc + a debug-build `debug_assert` against the ref.
- **Atomic + fallible history writes** *(Architect#6, RustSys#3, Safety#1/#3 — High).* `save_history` returns `Result` and writes a sibling temp file + `rename` (atomic on POSIX). Callers (`report`, `append_version_snapshot`) handle it; a write fault is now a per-version skip, not a process-aborting panic.
- **Worktree leak before guard** *(RustSys#4, Safety#2 — High/Medium).* `checkout` constructs the `CoreWorktree` RAII guard immediately after `git worktree add`, then fills sha/version — an error in `rev-parse` now unwinds through `Drop` instead of leaking the worktree + registry entry.
- **History pre-flight** *(API#4 — Medium).* A corrupt/unreadable history is a whole-run precondition: pre-flighted once in `main` (exit 2) before any rebuild, instead of failing identically after each expensive append. `MatrixError` doc now distinguishes abort-the-run vs per-version-recoverable.

### High-value optional (applied)
- Shared suite constants `DEFAULT_SIZES/DEFAULT_CONCURRENCY/BENCH_TILE_SIZE/BENCH_STREAMING_BUDGET` consumed by both `report` and `MatrixConfig::default` *(Architect#3, API#3)*.
- Pure `version_id` module extracted; `cross_version` repointed to it *(Architect#2)*.
- `CoreWorktree` fields made private + read accessors (Drop-safety) *(API#2)*.
- Typed `VersionOutcome::Skipped { error: MatrixError }` *(API#6)*; distinct exit codes 1=total/3=partial/2=usage+bad-history, documented *(RustSys#8, API#7)*.
- Owned target-dir RAII cleanup guard (panic-safe) *(Architect#9)*; TOML-escaped `paths` override *(RustSys#6, Safety#5)*; package-name pre-flight guard against a silently-ignored override *(Safety#4, RustSys#1)*.
- Real ignored **Rebuild-identity** e2e test + fast **Build-skip-continues** test *(Architect#5, Safety#7)* — replaces the old HEAD==HEAD-via-Reuse tautology.
- libvips feature forwarded to the per-tag build so the baseline cell is measured the same way (in-process FFI vs CLI) as the everyday report *(RustSys#2)*.
- Per-version heartbeat + wall-clock *(Architect#10)*; `remove()` reports success by end-state *(RustSys#9)*; `read_package_field` dedup of the two library scanner copies *(Architect#8, API#8)*.

### Conflicting recommendations + resolutions
1. **Shared vs per-tag target dir.** RustSys#1 → per-tag dirs / migrate to cargo `[patch]`; Architect documented sharing as a build-time optimization. **Resolved:** keep the shared dir (avoids full per-tag dep recompiles) and instead *detect* the risk it introduces — stale-exe removal + post-build self-report re-verification per tag + package-name guard + path escaping. `[patch]`/per-tag-dir migration declined as higher-effort with the detection now in place.
2. **When to persist.** Safety#3 offered accumulate-and-save-once (kills the O(N) rewrite) *or* atomic per-version save. **Resolved:** atomic per-version save — a many-minutes-per-tag runner benefits from partial results surviving a mid-sweep crash; the O(N) rewrite is negligible beside a full LTO build. Documented.
3. **History error class.** Global-precondition (API#4) vs per-version-skip (original). **Resolved:** both — pre-flight abort for a pre-existing corrupt file, per-version skip for a mid-run write fault.
4. **Fate of the Reuse variant.** Architect → `cfg(test)`/feature gate; API → rename + doc/gate. **Resolved:** rename + loud doc + `debug_assert` (`cfg(test)` is unusable — integration tests compile the lib without `cfg(test)`; a feature gate is heavier than warranted).

### Declined (low value / out of scope)
- **cargo `[patch]` migration / per-tag target dirs** *(RustSys#1)* — see resolution 1.
- **semver pre-release rank** *(RustSys#7)* — the axis records concrete builds; the test asserts `0.4.0-rc.1`≡`0.4.0`; caveat now documented.
- **Advisory lock/flock for concurrent writers** *(Safety#3)* — atomic rename applied; the runner is single-writer by design (documented).
- **SIGINT handler** *(Safety#6)* — `Drop` covers normal + panic; owned tree is RAII-removed; SIGKILL orphan is a documented `git worktree prune` recovery.
- **Bundle `(version, git_sha)` params into a struct** *(API#1)* — consistent ordering across the codebase; the sole matrix call site reads both from one `CoreWorktree`; latent footgun, not an active bug.
- **Rename `create_snapshot_for`/`VersionOutcome`; `#[non_exhaustive]`** *(API#9)* — house style; experts themselves advise skipping `non_exhaustive` absent external matchers.
- **`matrix` feature-gate for the whole runner** *(Architect#2 half)* — `publish=false` internal crate; the module extraction already inverts the coupling.
- **Aggregate colliding `version_key` rows** *(Architect#11)* — documented "same-key keeps first" in `lookup_row`; aggregation deferred.

### Recommended follow-up issues
1. **Prove the override linked the worktree.** The self-report reflects the `BENCH_CORE_DIR` stamp, so it can't independently detect a *silently-ignored* `paths` override where both sides trace to the worktree. A fully general check needs either `cargo metadata`/`--message-format=json` artifact-path resolution, or a runtime version symbol in the core crate (the latter is a change to `libviprs`, out of this session's scope).
2. **Bind-mount the core `.git` into the pinned Docker image** so the release-history axis can run in the controlled container instead of on the host *(Architect#4 option a)* — kept host-only for now, documented.

Issue #38, opened 2026-07-20
report: streaming/mapreduce timed cells panic (BudgetExceeded) at >=1024px under the default 1MB budget


spdrman:
## Summary

Running the `report` binary with its default config (`streaming_budget = 1_000_000`, sizes up to 4096x4096) makes the **streaming** and **mapreduce** timed child cells panic for every image >= 1024px, so those engines silently produce no rows at those sizes. Surfaced while smoke-testing the per-engine output-equivalence spot-check (#23).

## Evidence

`BENCH_ITERS=1 BENCH_WARMUP=0 cargo run --bin report` stderr:

```
called `Result::unwrap()` on an `Err` value: BudgetExceeded { strip_bytes: 1572864, budget_bytes: 1000000 }   # 1024
called `Result::unwrap()` on an `Err` value: BudgetExceeded { strip_bytes: 3145728, budget_bytes: 1000000 }   # 2048
called `Result::unwrap()` on an `Err` value: BudgetExceeded { strip_bytes: 6291456, budget_bytes: 1000000 }   # 4096
```

A single full-width strip (`width * strip_height * bpp`) exceeds the 1MB budget, and both `bench_streaming` and `bench_mapreduce` use `.with_budget_policy(BudgetPolicy::Error)` followed by `.run().unwrap()`. The child process crashes, `spawn_single_cell` returns `None`, and the engine is dropped from aggregation for that size. The executive verdict / comparison table at >=1024px therefore compare only `monolithic` vs `libvips`.

## Impact

- The headline comparison silently omits 2 of the 3 libviprs engines at every non-trivial size.
- Pre-existing; independent of #23. (The #23 spot-check inherits the gap — it can only score engines that produced a row — but does not cause it; its own candidate builder maps the same error to a graceful skip rather than a panic.)

## Options

1. Raise the default `streaming_budget` so a full-width strip fits at the largest configured size (budget must scale with width).
2. Make the two `bench_*` cells fail-soft (return `None` + eprintln on `EngineError`) so one unmeasurable config does not read as "engine absent" without explanation.
3. Auto-derive a per-size minimum budget from `compute_strip_height` / plan geometry.

Prefer (1)+(2): a realistic budget for the configured sizes, plus fail-soft so a genuinely-too-small budget degrades loudly instead of panicking.

PR #40, opened 2026-07-20
Output-equivalence: PSNR/SSIM mid-pyramid tile spot-check (#23)

### LINKED ISSUE #23: Output-equivalence: PSNR/SSIM tile spot-check
The per-level PNG tile-grid equivalence gate is in. Add a PSNR/SSIM spot-check of a few mid-pyramid tiles between libvips and each libviprs engine so a faster-but-visually-wrong result is flagged, not just a geometry match. (4-expert review, major.)

### LINKED ISSUE #32: output-equivalence: PSNR/SSIM tile spot-check gate
Sub-issue of #23. The equivalence gate checks per-level PNG tile GEOMETRY only. Add a PSNR/SSIM spot-check of a few mid-pyramid tiles between libvips dzsave and each libviprs engine on a committed fixture, so a faster-but-visually-wrong result is flagged (not just a geometry match). Test (fixture→libvips-differential): a fixture image → libvips dzsave + libviprs dzsave → PSNR of corresponding tiles >= threshold; a deliberately-corrupted tile trips the gate.

spdrman:
Resolves **#23** and sub-issue **#32**. Adds a PSNR/SSIM pixel spot-check of mid-pyramid tiles between libvips dzsave and **each** libviprs engine (per-engine, per-concurrency) on a committed fixture — so a fast-but-visually-wrong engine is flagged, not just a geometry match. Advisory metric `equivalence_psnr_db` surfaced in the report + serialized per row.

TDD: commit 1 failing tests (unit + vips-differential positive/negative); commit 2 implementation. 4-expert review applied (per-engine fix was the headline — consolidated below). Rebased onto #19/#24 (+ one integration commit initializing the new field in #19's version_matrix test). Local feature-matrix gate green (CI skipped via [skip ci]).

Closes #23
Closes #32


spdrman:
# Consolidated review — #23 (Output-equivalence: PSNR/SSIM mid-pyramid tile spot-check)

All findings triaged and the mandatory + high-value fixes applied in `8ba5176` (branch `bench/23-psnr`). Verification below.

## Overall assessment

The feature is well-conceived and unit-tested (self-contained PSNR/SSIM helpers + a real vips-differential positive/negative pair). The one thing that **had** to be fixed before merge is that all four reviewers independently converged on the same correctness defect, and it was flowing into **persisted** JSON. Around it sat a cluster of fail-soft and honesty issues.

- **Architecture** — sound; the pixel check is the right companion to the geometry gate.
- **Implementation quality** — good; a few public-surface honesty gaps (now closed).
- **Safety/reliability** — was the weak axis: a parent-process spot-check could panic and erase every measurement taken so far. Now fully fail-soft.
- **Performance** — irrelevant location (untimed, off the hot path); no changes needed.
- **Maintainability** — improved: honest names, typed skip reasons, codec-coupled discovery.
- **Production-readiness** — `benchmark_history.json` was carrying mislabeled per-engine numbers; that is the fix that gated merge.

## Mandatory fixes (applied)

### 1. [High ×3 — Systems, Safety, API] One monolithic@c0 PSNR was stamped onto every libviprs row
The headline finding, raised independently three times. `equivalence_psnr_db` is documented as "**THIS** engine's" but the score was measured once from a hardcoded `EngineKind::Monolithic` pyramid at concurrency 0, then written to the streaming, mapreduce **and** every concurrency>0 row — and it is serialized into `benchmark_results.json` / `benchmark_history.json` (consumed by cross_version + charts). Two of three rows persisted a fidelity number never measured from their own output.

**Conflicting recommendations + resolution.** Reviewers offered (a) measure per engine (and per concurrency), (b) prove the engines byte-identical then relabel as "libviprs-family", and (min) stamp monolithic only / others `None`. I took **(a), in full**: each libviprs engine now builds its **own** candidate pyramid at **the row's own concurrency** and is PSNR-compared against a shared per-size libvips reference; each row carries **only** its own score (`None` if that engine/config produced none). Chosen over (b)/(min) because streaming/mapreduce are exactly the "fast" engines a wrong-pixels bug would hide in, and it satisfies Safety's explicit "never borrow another config's number." Cost stays bounded — the reference is built once per size; candidates are untimed and reclaimed immediately (`EquivalenceReference`, `Drop`-cleaned).

### 2. [Medium — Systems & Safety] Parent-side spot-check could panic and discard the whole run
`write_libviprs_pyramid` (`.run().unwrap()`) and `write_temp_png` (`File::create().unwrap()`, encoder `.unwrap()`) ran in the parent before the child-isolated loop; any engine/fs error unwound the entire report. `write_libviprs_pyramid` and a new `write_png_at` now return `Result`; `EquivalenceReference` degrades every failure to a stderr note + no score. (`gradient_raster`'s `Raster::new().unwrap()` is left — a construction invariant over a buffer we just sized, identical to every other call site in the crate.)

### 3. [Medium — Safety / Low — Systems] Present-but-undecodable tile was skipped like a benign blank
`decode_rgb8` mapped any error to `None` and the loop `continue`d it, conflating "corrupt" with "legitimately-absent blank." New `read_tile` trichotomizes **Absent / Decoded / Corrupt**: a tile present on both sides but undecodable now scores 0 dB (trips the gate) instead of dropping out of the sample.

### 4. [Medium — API / Low — Systems] `spot_check` collapsed ≥3 outcomes into a bare `None`
Now returns `Result<TileFidelityCheck, SpotCheckSkip>` (`NoReferenceLevels` / `NoCandidateLevels` / `NoComparableMidLevel` / `NoCoPresentTiles`), each logged distinctly, so a structural divergence is never swallowed as the benign "image too small" case.

## High-value refinements (applied)

- **`ssim` → `global_ssim`** (Systems + API): the name now advertises a single global, channel-interleaved window (not windowed/per-channel MSSIM).
- **Race-free source PNG** (Systems): the equivalence source is written to `root/src.png` (unique per size), not the size-keyed path shared with the child cells; cleanup folds into the tree's `remove_dir_all`.
- **Report message honesty** (Safety): "libvips unavailable" now prints only when vips is truly absent; otherwise "no comparable mid level for the configured sizes."
- **"gate" → "advisory"** (Safety): the pixel check is non-fatal (loud WARNING + `[FAIL]`, no abort/exit); wording no longer overstates it.
- **Codec-coupled discovery** (Systems): `read_level_grids` derives its extension from `BENCH_TILE_SUFFIX` instead of hardcoding `png`.
- **Result struct cleanup** (API): `TilePsnrCheck` → `TileFidelityCheck`; dropped the 3 never-read grid fields; per-field docs; removed the float-eq `PartialEq` derive.
- **Hardening** (Safety): `u64` tile count (overflow parity with the sibling filter) + `debug_assert!(min_psnr.is_finite())` guarding the serde path; report's avoidable `key.clone()` removed.

## Deliberately declined / partial (with reasons)

- **write_libviprs vs write_libvips tile-size currency (API Medium) — partial.** Mismatch risk eliminated on the measured path: `EquivalenceReference` is the single orchestrator and derives both sides from one `tile_size`. Writers kept public (the negative test builds → corrupts → re-checks and needs them separately) rather than collapsed to one entry point.
- **psnr/ssim `0.0` sentinel → `Option` (API Low) — partial.** Kept the sentinel (it already trips the gate; the spot-check only ever feeds equal-length tiles) and promoted the collision to a prominent doc note — the reviewer's stated fallback. `Option<f64>` would ripple through the hot loop and every test for no behavioral gain.
- **`#[non_exhaustive]` on the result struct — declined.** `publish = false` internal crate, no external constructors.
- **Persist mean/SSIM sub-struct (API) — declined.** Schema bump + history migration is out of scope; min-PSNR is the gate. Documented "min-only, mean/SSIM stderr-only" in the field doc (their fallback).
- **Integer accumulation for psnr/ssim (Systems) — declined.** Reviewer said leave as-is; textbook f64 form, cost irrelevant here.
- **vips subprocess timeout (Safety) — declined.** No `wait_timeout` in std; the parent vips is the same binary the suite already invokes per timed cell.
- **Unify `per_level_png_tiles` / `read_level_grids` enumerators (API) — declined.** Different return shapes in different modules; fixed the concrete coupling (codec extension) instead of the larger refactor.
- **Architect "[High] test: test → test" — no actionable content** (placeholder).

## Note surfaced by the end-to-end smoke (pre-existing, out of scope) → filed as #38
Running `report` shows the **streaming/mapreduce timed child cells panic** (`BudgetExceeded`) at ≥1024px under the default 1 MB budget — `bench_streaming`/`bench_mapreduce` still `.run().unwrap()` — so those rows never reach aggregation and get no spot-check. Independent of #23 (the #23 candidate builder maps the same error to a graceful skip). Filed as **libviprs/libviprs-bench#38**.

## Verification
`cargo build --all-targets` ✓ · `cargo fmt --check` ✓ · `cargo clippy --all-targets -- -D warnings` ✓ (only Cargo's pre-existing pdfium `[patch]`-unused notice, not a lint) · `cargo test` **27/27** ✓ (output_equivalence 6/6, incl. vips-differential positive + negative). Smoke (`report`, `BENCH_ITERS=1`): per-engine PSNR lines render; monolithic 100.0 dB min/mean, SSIM 1.0000 at 1024/2048/4096; streaming+mapreduce built and correctly skipped at 512 (no multi-tile mid level); each row's `equivalence_psnr_db` is its own score. `8ba5176`.

PR #41, opened 2026-07-20
Real-content PDF fixture + rasterized-PDF scalability workload (#22)

### LINKED ISSUE #22: Committed real-content fixture (rasterized PDF page) for the scalability workload
Chart titles now honestly say 'synthetic gradient'. Add a real-content workload: rasterize a committed fixture PDF via the pdfium feature (or commit a rasterized page) at increasing sizes, reported as a separate chart series, so the comparison isn't only a synthetic gradient. (4-expert review, major.)

### LINKED ISSUE #30: fixtures: commit a small real PDF page fixture
Sub-issue of #22. fixtures/ currently holds only .DS_Store. Add a small, freely-licensed real PDF page (fetch from the internet, checksum-verified, committed with provenance) to serve as a real-content benchmark workload — replacing the synthetic-gradient-only story.

### LINKED ISSUE #31: scalability: rasterized-PDF workload series via pdfium
Sub-issue of #22. Add a scalability workload variant that rasterizes the committed PDF fixture (via the existing pdfium feature / PdfiumStripSource) at increasing sizes, reported as a separate chart series alongside the synthetic gradient — so the libvips-vs-libviprs comparison covers real content. Test: the fixture rasterizes to expected dimensions and the workload runs end-to-end at a small size.

spdrman:
Resolves **#22** and sub-issues **#30** (committed CC0 PDF fixture, SHA-256 pinned in PROVENANCE.md) and **#31** (feature-gated rasterized-PDF scalability series via PdfiumStripSource, plotted alongside the synthetic-gradient engines). Chart titles/footnotes disclose the workload; a `--pdf-max-mp` cap bounds the heavy PDF cell; typed `PdfBenchError` replaces stringly failures.

TDD: commit 1 fixture + failing tests; commit 2 implementation. 4-expert review applied (consolidated below). Rebased onto #19/#23/#24 (lib.rs two-function collision + new equivalence field reconciled). pdfium stays behind its feature gate (default build byte-identical). Local feature-matrix gate + pdfium clippy green (CI skipped via [skip ci]).

Closes #22
Closes #30
Closes #31


spdrman:
## Consolidated review — #22 (real-content PDF fixture + rasterized-PDF scalability workload)

Four expert reviews (Principal Architect, Rust Systems, Safety & Reliability, API Design). Findings are de-duplicated below. All mandatory fixes and the high-value refinements are applied in `6e5ca16` (`refactor: apply 4-expert review (#22) [skip ci]`).

### Overall assessment

The feature is well-scoped and cleanly gated: a checksum-pinned, CC0, single-page real-content fixture drives a fifth `streaming-pdf` series, feature-gated so the default build is byte-for-byte unchanged. The provenance/integrity discipline (SHA-256 enforced in-test *and* in `PROVENANCE.md`) is a genuine highlight.

The dominant risk the panel converged on is **honesty of the comparison, not correctness**. The PDF line is plotted on the same axes as the four gradient engines but is confounded on *every* metric: its memory number is a shared in-process RSS high-water mark (not per-run), and its time-based numbers include pdfium vector rasterization the gradient series never pay. Three of the four experts independently flagged the RSS confound. Secondary issues: a temp-dir leak on the error path, a stringly-typed error that hid real regressions behind the "pdfium unavailable" skip, a fixture-decoupled magic constant, and an uncapped sweep rendering a 65 KB vector page at up to ~1210 DPI.

| Dimension | Assessment |
|---|---|
| Architecture | Sound — gating and series separation are correct. One structural suggestion declined (see Conflicts). |
| Implementation | Good — main gaps were error typing and cleanup, now fixed. |
| Safety/reliability | Temp-dir leak + overflow-prone budget math fixed; typed error separates environment-skip from regression. |
| Performance | Sweep capped: the PDF series now stops at 50 MP instead of upsampling to 280 MP / ~1210 DPI. |
| Maintainability | Magic constant removed (derived from the fixture); source-level test tightened to the wiring, not the prose. |
| Production-readiness | It's a benchmark, not shipped code — with the disclosures it is safe to publish. |

### Mandatory fixes (applied)

1. **Typed error + temp-dir leak** (Architect M4, Safety L2). New `PdfBenchError { SourceUnavailable(PdfError), Plan(PlannerError), Engine(EngineError) }`. `run_pdf_streaming` now skips **only** `SourceUnavailable` (the libpdfium-unavailable case) and **panics** on `Plan`/`Engine` — matching the gradient runners' `.unwrap()` — so a real planner/engine regression can no longer vanish silently. The temp dir is reclaimed on **every** exit (capture result → always `remove_dir_all` → then propagate).
2. **RSS confound disclosed** (Architect M1, RustSys M2, Safety M1). `getrusage(RUSAGE_SELF).ru_maxrss` is a process-wide monotonic high-water mark, and the PDF cell runs last (after the monolithic ~1 GB canvas), so its peak-RSS and the efficiency/resource-cost derived from it are not per-run. Disclosed via a per-metric **SVG footnote** (memory chart → "RSS is a shared high-water mark, see `tracked_memory_mb` in the JSON"), a console caveat block, the chart NOTE, and a softened `bench_streaming_pdf` docstring (the bounded figure is the engine's *tracked* memory, not the process RSS).
3. **Rasterization-in-timer disclosed** (RustSys M1). The gradient sources are materialised before the timer; the PDF rasterizes inside the timed run. The four time-based charts now footnote "end-to-end rasterize+pyramid (RGBA, serialized pdfium); the gradient series are pyramid-only."
4. **Chart labelling** (Architect M2). `{workload}` added to the efficiency and resource-cost titles (previously only charts 1–3 named the workload); the RGBA / larger-budget / serialized-render confound is disclosed alongside.
5. **Sweep cap** (Architect M3). `DEFAULT_PDF_MAX_MP = 50` + `--pdf-max-mp <n>`; the PDF companion is capped while the gradient series stay uncapped.
6. **Magic constant removed** (Architect L5, RustSys L4, Safety L4). `pdf_base_width()` reads the fixture's real 72-DPI width once at startup; `pdf_dpi_for_width(target, base_width)` scales from that, so a PROVENANCE-permitted fixture swap stays correct automatically.

### High-value refinements (applied)

- **Saturating budget math** (Safety) — `saturating_mul` throughout, so adversarial `dpi`/`tile_size` can't wrap the budget small or panic in debug.
- **Graceful test skip** (Safety) — the two `#[cfg(pdfium)]` tests now skip (not cryptic dlsym panic) **only** when libpdfium can't bind; all dimension/non-blank/pyramid assertions run post-open, so real regressions still fail hard where a compatible pdfium exists.
- **Explicit budget ownership** (RustSys/Safety) — `run_pdf_streaming` passes `STREAMING_BUDGET_FLOOR`; `bench_streaming_pdf` owns the RGBA budget (removes the dead-input smell).
- **Tightened source-level test** (Architect/RustSys suggestion) — new guard asserts the call site *and* `all_points.push(p)` (neither appears in a comment), so deleting the wiring while keeping the prose no longer stays green.
- **Cargo.toml sha2 comment** (RustSys L4) — corrected: the pin matches libviprs's direct `sha2 0.11`, but the graph is **not** single-version (lopdf still pulls a 0.10.x transitively). Confirmed with `cargo tree -i`.

### Conflicting recommendations + resolution

- **RustSys 1 vs RustSys 2 / Safety 1.** RustSys 1 wants the PDF series restricted to the peak-memory chart and omitted from the time charts; RustSys 2 / Safety 1 show the peak-memory (RSS) chart is exactly where the PDF series is *most* misleading. These are directly opposed. **Resolution:** keep the PDF series on all five charts with **per-metric disclosure** (the memory footnote redirects to `tracked_memory_mb`; the time footnote flags rasterize+pyramid), rather than concentrating it onto the single most-misleading chart.
- **Architect 1 (slim observer-free runner).** Declined the restructure. Because RSS is already pinned ≥ 1 GB by the monolithic canvas earlier in each size iteration, the observer's O(tiles) Vec cannot move the high-water mark — the stated RSS benefit is moot — and keeping one `bench_streaming_pdf` path preserves the single observer-backed path the integration test asserts `strips > 0` through. The real honesty gap is addressed by disclosure, which a slim runner would not fix.

### Declined / deferred (with reason)

- **De-duplicate `bench_streaming_pdf` vs `bench_streaming`** (RustSys L3, Low) — declined to keep the diff focused on correctness/honesty and avoid churning the stable, widely-used `bench_streaming`; **recommend a standalone cleanup issue.**
- **Run PDF series once (c=1)** (RustSys) — declined (asymmetric across the two thread-budget chart sets); the concurrency-axis caveat is in the disclosure.
- **Subprocess-isolated per-run PDF RSS** — the fully-correct fix for the RSS confound; deferred as an enhancement (**recommend an issue**).
- **API & Library Design Expert** — the submitted finding is an empty placeholder (`src/lib.rs:549`, "test" → "test") with no substantive content; nothing to apply.

### Verification

`cargo build` (default + `--features pdfium`), `cargo fmt --check`, `cargo clippy --all-targets -D warnings` (default + pdfium), and `cargo test` (default + pdfium) all green. Smoke: `scalability --replot` over a synthetic dataset with a `streaming-pdf` series renders 5 SVGs carrying the `{workload}` titles and per-metric footnotes; the gradient-only replot is unchanged (title "synthetic gradient", 450 px, no footnote). The live pdfium render path can't execute here (the bundled libpdfium predates a symbol the pinned pdfium-render fork imports), so the pdfium-gated logic is covered by build+clippy and the graceful-skip test path rather than a live render.

Issue #43, opened 2026-07-20
charts: restore large-image x-axis zoom in the JS scalability renderer (supersedes --crop)


spdrman:
Follow-up from the #20/#21 charting rework (branch `bench/charts-js-lib`).

The rework removed the scalability x-axis crop capability from `src/scalability.rs` (`DEFAULT_X_AXIS_CROP_MP = 15`, `--crop`/`--no-crop`, and `--replot`) with no JS replacement for the zoom.

- `--replot` is well superseded: `render.mjs` always re-renders from JSON, a strict improvement.
- `--crop` is largely obsoleted: log-log axes inherently spread the small-image regime the crop existed to hide.
- But the large-image zoom (focus the axis on the >=15 MP tail) has no equivalent — only `--linear` exists.

If the large-image zoom is still wanted, add an optional `xMin`/domain bound to `renderScalabilityChart` (it already threads bounds through `log10Scale`/`decadeTicks`/`enclosingDecades`) and a `--min-mp` flag to `render.mjs` that filters points and recomputes the domain. Otherwise, close as 'log-log supersedes it'.

Issue #44, opened 2026-07-20
charts: golden round-trip test to catch Rust↔JS producer/consumer field-shape drift


spdrman:
Follow-up from the #20/#21 charting rework (branch `bench/charts-js-lib`).

`render.mjs` consumes the harness JSON by reading named fields (`wall_time`, `peak_rss_bytes`, `megapixels`, `wall_time_ms`, …) and coercing anything missing to a default/NaN (then dropping the point). If the Rust structs evolve (e.g. `wall_time` → `wall_time_ns`, a `peak_rss_bytes`/`resource_cost` rename), `render.mjs` reads valid JSON but yields zero usable points.

Already landed in the rework (this review):
- `render.mjs` now distinguishes 'file absent' from 'present + parsed but 0 charts', and its `historyShapeOk`/`scalabilityShapeOk` probes emit a LOUD stderr + non-zero exit when records are present but lack the load-bearing fields (likely field-shape drift), vs a quiet informational note when there is simply not enough data yet (e.g. a single snapshot).

Still wanted here:
- A committed GOLDEN emitted by the Rust harness itself (not the hand-authored fixtures, which freeze today's shape by hand) that a CI/pre-commit check round-trips through `render.mjs`, so producer/consumer drift is caught mechanically. The hand-authored `tools/charts/fixtures/*.json` should stay, but a harness-emitted golden closes the 'fixture drifted from the real struct' gap.

Issue #42, opened 2026-07-20
charts: finish the JS migration — port grouped-bar chart_*.svg to render.mjs and drop plotters


spdrman:
Deferred half of the #20/#21 charting rework (branch `bench/charts-js-lib`).

The history-trend and scalability SVGs now render in JS (`tools/charts/*.mjs`) from the harness JSON, but the six grouped-bar comparison charts (`chart_wall_time.svg`, `chart_peak_memory.svg`, `chart_throughput.svg`, `chart_efficiency.svg`, `chart_resource_cost.svg`, plus the savings bar) are still emitted in Rust via `libviprs_bench::generate_charts` (plotters). Consequences:

- `plotters` remains a hard dependency (`Cargo.toml` + `use plotters::prelude::*` in `src/lib.rs`).
- The engine palette is duplicated across languages: `COLOR_VIPS/COLOR_MONO/COLOR_STREAM/COLOR_MR` in `src/lib.rs` mirrors `COLORS` in `tools/charts/chart.mjs` and the two must be hand-kept in lockstep (documented as such for now).

Scope:
- Port the grouped-bar renderer to `chart.mjs` (causl-bench's `renderMetricGroupedBars` is the template) and have `report` emit its per-run metrics as JSON (it already writes `benchmark_results.json`).
- Render all `chart_*.svg` via `render.mjs`; delete `generate_charts` + the Rust palette; drop the `plotters` dependency.
- Then the engine palette has a single source of truth (the JS `ENGINES` array).

PR #45, opened 2026-07-20
Adopt causl-bench JS SVG chart code: history-trend + scalability log-log (#20, #21)

### LINKED ISSUE #20: History-trend chart series alignment + version ordering
generate_history_chart still indexes points positionally per-engine, misaligning series when an engine is missing from a snapshot — index points by snapshot position and gap missing engines. Also cross_version orders versions lexicographically — use semver/timestamp and key duplicate versions by version@short_sha. (4-expert review, major.)

### LINKED ISSUE #21: Scalability charts: true log-log axes
Full-range linear is now the default (15-MP crop opt-in via --crop). Implement true log-log axes so both the small-image and large-image regimes are readable on one chart without cropping the regime where libvips looks best. (4-expert review, major.)

### LINKED ISSUE #28: history-chart: index series by snapshot position + gap missing engines
Sub-issue of #20. generate_history_chart indexes points positionally within each per-engine vector, misaligning series against version ticks when an engine is missing from a snapshot. Index every point by its snapshot position in the history (x=snapshot index) and draw gaps (break the line) where an engine has no run, so series stay aligned. Test: a history where one snapshot omits an engine renders that engine's points at the correct x with a gap, not compacted.

### LINKED ISSUE #29: scalability: true log-log axes
Sub-issue of #21. Scalability charts currently use full-range LINEAR axes (15-MP crop opt-in). Implement true LOG-LOG axes so both the small-image and large-image regimes are readable on one chart without cropping the regime where libvips looks best. Keep linear as an opt-in. Test: the log axis maps points across ~1MP..280MP to sensible log positions with log-spaced ticks; a decade spans equal pixels.

### LINKED ISSUE #34: charts: vendor the causl-bench SVG chart code into tools/charts + wire run-bench.sh
Shared infrastructure for #20 and #21. Port the proven, dependency-free SVG chart renderers from causl-bench (packages/bench/src/chart.ts: renderHistoryTrend, renderScalabilitySweep + the COLORS/formatNumber/svgPlaceholder helpers) into libviprs-bench as tools/charts/ (plain ESM .mjs, node --test — matching libviprs-org/cli/tools style, no TS build). Adapt the series model to libviprs's engines (monolithic/streaming/mapreduce + the libvips oracle) and megapixel scales. Add a render CLI that reads the JSON the Rust harness already emits (benchmark_history.json + the scalability results) and writes the SVGs; wire run-bench.sh to invoke it after the Rust harness. REMOVE the now-duplicated Rust SVG emitters (lib.rs generate_history_chart, scalability.rs render_charts) while KEEPING the JSON-producing data path + its Rust tests. Gives visual + code consistency with the causl-bench article.

spdrman:
Resolves **#20**, **#21** and sub-issues **#28**, **#29**, **#34**. Ports causl-bench's proven, dependency-free SVG chart renderers into `tools/charts/*.mjs` (plain ESM + node --test), driven by the JSON the Rust harness already emits, wired into run-bench.sh — and removes the now-duplicated Rust plotters emitters (`generate_history_chart`, `render_charts` + helpers), keeping the data path + its Rust tests.

- **#20/#28**: history-trend chart indexes by SNAPSHOT INDEX with version ticks + BREAKS the polyline for a missing engine (gap, not compaction).
- **#21/#29**: scalability chart is true LOG-LOG (log10 both axes, decade ticks); linear preserved via `--linear`.
- **#34**: shared chart code in `tools/charts/` (colours/engine-order match the prior Rust palette), consistent with the causl-bench article.

TDD: commit 1 failing JS tests + fixtures; commit 2 port + Rust-emitter removal; +4-expert review applied (37 JS tests). Rebased onto #19/#22/#23/#24 (resolved the lib.rs/report.rs/scalability.rs overlaps — kept #19's atomic-save match + #22's PDF data-gen, removed only the chart emitter). **CI job the review added was dropped** — repo gates locally. Filed follow-up #43 (optional large-image zoom). Local feature-matrix gate + 37 JS tests green (GitHub CI skipped via [skip ci]).

Closes #20
Closes #21
Closes #28
Closes #29
Closes #34


spdrman:
## Consolidated 4-expert review — charts JS-lib rework (#20/#21) — applied

Reviewed the committed rework (JSON-in-Rust ↔ SVG-in-JS) across all four lenses. Findings de-duplicated, conflicts resolved, and every mandatory + high-value fix applied on `bench/charts-js-lib` (**`ba98d4a`**, `[skip ci]`). Verification all green (see bottom). Deferred work filed as issues, not dropped.

### Overall assessment
- **Architecture** — sound direction, but the migration was *two-thirds done*: history + scalability moved to JS while the six grouped-bar `chart_*.svg` stayed in Rust/plotters, leaving the engine palette duplicated across languages. Resolved for now by **scoping the split explicitly** (documented in `lib.rs` + `chart.mjs`, single source of truth on the JS side) and filing **#42** to finish the port and drop plotters.
- **JS correctness** — solid port, but the log-log axis math had real defects: sub-decade ranges rendered a **label-free axis**, decade labels **collided** (`0.001`→`0`), and `Math.max(...arr)` would **throw on large histories**. All fixed.
- **Safety / determinism** — the most important cluster: log mode **silently dropped ≤0/failed points, bridged them with a straight line, and made an all-failed engine vanish from the legend**; `render.mjs` **conflated "file absent" with "shape drift"**. All fixed, with the previously-untested degenerate cases now pinned.
- **API / maintainability** — engine identity was three hand-synced structures with a **silent-drop** failure mode; `ENGINE_LABELS` was dead + a fidelity drift. Unified to one frozen source of truth; labels wired back in.
- **Fidelity to causl** — high and now higher: legends restore the title-cased `Monolithic/Streaming/MapReduce` the Rust artifacts used; a docstring records where the libviprs shape intentionally diverges (`renderScalabilityChart` vs causl's `renderScalabilitySweep`).

### Conflicting recommendations + resolutions
1. **`peak_rss` serde-alias (Architect L) vs treat 0 as missing (Safety S)** — for *history*, the Architect's `peak_rss_bytes ?? peak_memory_bytes` is **incorrect**: in the Rust structs `#[serde(alias="peak_memory_bytes")]` sits on `tracked_memory_bytes` (engine-tracked working set), *not* `peak_rss_bytes`; the removed emitter read only `peak_rss_bytes` (0 for legacy). Mirroring that alias would conflate the two memory bases #153 deliberately separated. **Adopted Safety's "0/absent ⇒ missing (NaN)"** for the history memory metric. The alias *is* real for *scalability* (`peak_rss_mb` genuinely aliases `peak_memory_mb`), so **applied there**.
2. **Enclosing-decade snap (API M) vs fallback endpoint labels (Architect/Systems M)** — same root cause, two proposed fixes. The **domain snap** (`[10^⌊log⌋, 10^⌈log⌉]`) *subsumes* the fallback: enclosing decades always yield ≥2 labeled majors and give the extreme dots headroom off the frame edge. Implemented the snap alone; no separate min/max annotation needed.
3. **Render non-fatal (API/Safety S) vs exit-nonzero on drift (Safety M)** — not actually in conflict. `render.mjs` exits non-zero on *shape drift*; `run-bench.sh` wraps the call in `|| echo …` so the measurement JSON always survives a chart-side failure. Both satisfied.
4. **Skip single-point polylines (Systems L)** — **declined**: the committed #20 gap test relies on the isolated run-3 streaming point emitting a segment, and a 1-point polyline is visually inert; skipping it would desync the two renderers and force rewriting a deliberate regression test.

### Mandatory fixes (all applied)
| Fix | Sources | What changed |
|---|---|---|
| Sub-decade axis had **no labels** | Arch M / Sys M / API M | `enclosingDecades()` snaps the log domain; drives `log10Scale` + `decadeTicks`. Efficiency (3‑4) now labels 1/10/100. |
| Decade-label **collision** `0.001`→`0` | Sys M | New `formatLogTick()` (compact `1e-3` for tiny/huge, trimmed decimal otherwise); linear mode keeps `formatNumber`. |
| Log mode **bridges dropped points + engine vanishes** | Safety M | Break the polyline at any ≤0/non-finite size (never interpolate across it); legend keeps every engine with *input* points; visible "N pts ≤0 omitted" note. |
| Degenerate/failure cases **untested** | Safety M | Added tests: ≤0/NaN/Inf mix, zero-span, single-point, single-snapshot, all-non-finite, `log10Scale`/`decadeTicks` guards. |
| `render.mjs` **conflates absent vs drift** | Safety M | `historyShapeOk`/`scalabilityShapeOk` field probes → loud stderr + non-zero exit on drift, quiet informational note on "not enough data yet". |
| Engine identity split + **silent drop** of a non-canonical engine | API M | One frozen `ENGINES` source of truth derives `ENGINE_ORDER`/`COLORS`/`ENGINE_LABELS`; renderers draw canonical order then append data-only engines with a fallback colour. |
| Node tests in **no gate** | Arch M / API L | `tools/charts/package.json` (`type:module`, `test` script) + a `charts` CI job (`node --test '*.test.mjs'`; the dir form is broken on current Node). |

### Optional refinements applied (high value, low effort)
Gap logic keyed on the drawn-snapshot ordering so a wholly-absent snapshot no longer shatters every line (Sys L) · reduce-based `minOf`/`maxOf` (no `Math.max(...arr)` overflow) (Sys L) · `escapeXml()` on all interpolated text (Sys S) · `log10Scale` guarded against non-positive domain (Safety S) · scalability `peak_rss_mb ?? peak_memory_mb` alias (Arch L) · `Object.freeze` palette/order + `ENGINE_LABELS` wired into both legends, tests relaxed case-insensitively (API L) · shared present-engine legend convention (API L) · `run-bench.sh` render made non-fatal (API/Safety S) · README Charts section + Node prerequisite, scalability size-grid comment de-crop-ified, `lib.rs` palette-mirror note (Arch L / Safety L) · `history_migration` test re-pointed from the dead label-prefix filter to the structured `width/height/concurrency` grouping key (Safety/API L) · `render.mjs` value-flag validation + real `--help` (API L) · naming cross-ref docstring (API S).

### Declined (low value — one-line reason)
- **`historyPointsFor` O(configs²) micro-opt** (Sys S) — premature; histories are small.
- **Re-add argv rejection to the `scalability` binary** (Sys L) — took the reviewer-sanctioned "or document" path (module doc + README state direct runs emit JSON only; `render.mjs` is the replot); avoids breaking wrappers passing the old `--no-crop` no-ops.
- **Binaries shell out to `render.mjs`** (Arch L) — `run-bench.sh` is the documented chart entrypoint; keeps a clean Rust↔node boundary.

### Deferred to issues (repo "defer work to issues" norm)
- **#42** — finish the JS migration: port grouped-bar `chart_*.svg`, drop `plotters`, collapse the cross-language palette.
- **#43** — restore large-image x-axis zoom in the JS scalability renderer (log-log supersedes `--crop` for the small-image regime; the large-image zoom has no replacement).
- **#44** — committed harness-emitted golden round-tripped through `render.mjs` to catch producer/consumer field-shape drift mechanically.

### Verification
`node --test`: **37 pass / 0 fail** · `cargo fmt --check`: clean · `cargo clippy --all-targets --features libvips -- -D warnings`: clean · `cargo test --lib --tests --doc --features libvips`: all green (lib 10 + integration + doc) · `render.mjs` smoke: 12 SVGs, zero `NaN`/`undefined`, sub-decade axes now labeled.

Issue #47, opened 2026-07-20
bench: streaming budget admission is reverse-engineered from the core engine's private pre-flight — adopt a core-exposed primitive


spdrman:
Follow-up from the #38 four-expert review (architecture/API finding, deferred; needs a core-crate change to resolve cleanly).

## Problem
`streaming_budget_for` (src/lib.rs) re-derives the streaming/mapreduce engines' *private* budget-admission formula — `canvas_width * (2 * tile_size) * bpp` — which lives verbatim in two core locations (`libviprs/src/streaming.rs` and `libviprs/src/streaming_mapreduce.rs`). That admission threshold is an internal invariant the core does not expose as API, so the bench hard-codes a third copy of it and layers its own headroom policy (`floor.max(.)`, `2x`) on top.

The core already models the full per-level working set exactly (`estimate_streaming_memory` -> `PyramidPlan::estimate_streaming_peak_memory`) and exposes the forward direction `compute_strip_height(plan, format, budget) -> Option<u32>`, but NOT the inverse — the minimum budget that survives pre-flight — so the bench reverse-engineers it.

## Direction
- Ideally the core exposes a canonical primitive, e.g. `libviprs::minimum_streaming_budget(plan, format) -> u64` (the smallest budget that passes the pre-flight gate — the inverse of `compute_strip_height`), and BOTH the engines' own pre-flight and this bench call it, so the admission threshold has one definition.
- The bench's headroom policy (the `floor.max(.)` and the margin) then layers on top of that primitive, and `streaming_budget_for` stops restating the `2 * tile_size` rule as a bare literal.
- Once that lands, `streaming_budget_for`'s body becomes `floor.max(minimum_streaming_budget(plan, format).saturating_mul(k))` (or is sized directly from `estimate_streaming_peak_memory`), deriving the rule from the engine instead of duplicating it.

Tracked on the bench side because that is where the coupling bites; the enabling change is a core-crate API addition.

Issue #46, opened 2026-07-20
bench: bench_streaming/bench_mapreduce still .run().unwrap() — a non-budget engine error panics the whole report and leaks a temp dir


spdrman:
Follow-up from the #38 four-expert review (reliability finding, deferred as out of scope for the budget-sizing fix).

## Problem
`bench_streaming` and `bench_mapreduce` (src/lib.rs) still end with `.run().unwrap()`. #38 removed the *budget-driven* trigger of that panic, but the fragile pattern remains: any OTHER engine error (sink IO failure, source error, downscale error) still panics and drops every streaming/mapreduce row from the report — the same report-wide-failure mode #38 was about.

Additionally, the panic fires before the `out_dir` cleanup (`remove_dir_all`), so on that path a temp directory is leaked under `$TMPDIR`.

## Direction
- Make `bench_streaming` / `bench_mapreduce` return `Result` (or record a failed-row marker) instead of unwrapping, so a single engine error degrades to a missing/failed row rather than aborting the whole sweep.
- Move the `out_dir` cleanup so it also runs on the error path, mirroring `bench_streaming_pdf`'s reclaim-on-every-exit (it already computes the tile counts and `remove_dir_all`s before consuming the run result).

Non-blocking; the timed paths only panic on genuine engine faults today, but the report should surface those as a failed row, not lose the whole engine series.

PR #48, opened 2026-07-20
report: size streaming/mapreduce budget so large cells don't panic (#38)

### LINKED ISSUE #38: report: streaming/mapreduce timed cells panic (BudgetExceeded) at >=1024px under the default 1MB budget

spdrman:
Resolves **#38**. The report harness drove streaming/mapreduce with a flat ~1 MB budget under `BudgetPolicy::Error`, so from 1024px up the worst-case tile-aligned strip exceeded the budget → `BudgetExceeded` panic → streaming/mapreduce rows silently vanished at large sizes (only monolithic/libvips survived). Fix: a shared `streaming_budget_for(floor, canvas_width, tile_size, bpp)` helper sizes the budget to admit the worst-case strip (saturating math), wired into bench_streaming/bench_mapreduce/bench_streaming_pdf + write_libviprs_pyramid + flamegraph.rs; scalability's duplicate copy deleted (single source). report + snapshot now report the effective per-run budget.

TDD: commit 1 RED tests (reproduce the panic at 1024²); commit 2 fix. 4-expert review applied. Rebased onto charts/#22/#23/#24 (removed the crop const + scalability's duplicate helper). Filed follow-ups #46 (.run().unwrap() → Result) and #47 (core-exposed budget primitive) — **both to be resolved in the convergence sweep**. Local gate green.

Closes #38

### DIFF
--- a/src/flamegraph.rs
+++ b/src/flamegraph.rs
@@ -20,7 +20,7 @@ use libviprs::{
     CollectingObserver, EngineBuilder, EngineConfig, EngineEvent, EngineKind, Layout, MemorySink,
     PyramidPlanner, RasterStripSource,
 };
-use libviprs_bench::gradient_raster;
+use libviprs_bench::{gradient_raster, streaming_budget_for};
 use std::sync::Arc;
 
 const WIDTH: u32 = 4096;
@@ -126,6 +126,13 @@ fn main() {
     let planner = PyramidPlanner::new(WIDTH, HEIGHT, TILE_SIZE, 0, Layout::DeepZoom).unwrap();
     let plan = planner.plan();
 
+    // Size the streaming/mapreduce budget to admit the worst-case tile-aligned
+    // strip at this 4096-wide canvas: a flat 1 MB trips `BudgetPolicy::Error`'s
+    // `BudgetExceeded` up front (worst-case strip = 4096·512·3 = 6.29 MB) and
+    // `.run().unwrap()` would panic — the exact failure mode issue #38 fixes.
+    // RGB8 gradient, so bpp = 3; shared with the report/scalability paths.
+    let budget = streaming_budget_for(1_000_000, plan.canvas_width, TILE_SIZE, 3);
+
     println!("Image: {WIDTH}x{HEIGHT}, tile_size={TILE_SIZE}");
     println!(
         "Plan: {} levels, {} tiles",
@@ -175,7 +182,7 @@ fn main() {
         let result = EngineBuilder::new(strip_src, plan.clone(), &sink)
             .with_engine(EngineKind::Streaming)
             .with_config(EngineConfig::default())
-            .with_memory_budget(1_000_000)
+            .with_memory_budget(budget)
             .with_budget_policy(BudgetPolicy::Error)
             .with_observer_arc(observer.clone())
             .run()
@@ -216,7 +223,7 @@ fn main() {
         let result = EngineBuilder::new(strip_src, plan.clone(), &sink)
             .with_engine(EngineKind::MapReduce)
             .with_config(EngineConfig::default().with_concurrency(0))
-            .with_memory_budget(1_000_000)
+            .with_memory_budget(budget)
             .with_budget_policy(BudgetPolicy::Error)
             .with_observer_arc(observer.clone())
             .run()
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -83,7 +83,12 @@ pub struct BenchmarkSnapshot {
     pub timestamp: String,
     /// Tile size used.
     pub tile_size: u32,
-    /// Memory budget used for streaming/mapreduce engines.
+    /// Requested streaming/mapreduce budget FLOOR (the report's configured
+    /// [`BENCH_STREAMING_BUDGET`]), not the budget any single run actually used.
+    /// The effective budget is sized up per canvas width by
+    /// [`streaming_budget_for`] (issue #38), so cross-size rows are NOT under one
+    /// identical budget; each run's effective figure lives in
+    /// [`RunMetrics::memory_budget_bytes`].
     pub memory_budget_bytes: u64,
     /// Individual run metrics.
     pub runs: Vec<RunMetrics>,
@@ -180,7 +185,12 @@ pub struct RunMetrics {
     pub inflight_strips: u32,
     /// Concurrency level used.
     pub concurrency: usize,
-    /// Memory budget in bytes (streaming/mapreduce only, 0 for monolithic).
+    /// EFFECTIVE memory budget the engine actually ran with, in bytes
+    /// (streaming/mapreduce only, 0 for monolithic). This is the per-run figure
+    /// after [`streaming_budget_for`] raised the requested floor to admit the
+    /// worst-case tile-aligned strip at this canvas width (issue #38), so it is
+    /// size-dependent and generally exceeds the report's flat floor (the
+    /// [`BenchmarkSnapshot`] `memory_budget_bytes`).
     pub memory_budget_bytes: u64,
 }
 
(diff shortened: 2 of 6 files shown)


spdrman:
# Consolidated review — apply the 4-expert review of #38 (streaming/mapreduce budget sizing)

The #38 fix (size the budget so large cells don't panic with `BudgetExceeded`) is correct in mechanism, but the four reviews converge on the same structural gaps: it sized from the **wrong width**, left **three sibling paths still exposed to the exact panic it fixed**, **reported a budget no size used**, and **documented a 2× headroom the arithmetic contradicts**. All are addressed; two genuinely core-side / reliability items are deferred to issues per the repo's defer-to-issues convention.

## Overall assessment
- **Architecture**: the budget-admission formula (`canvas_width × 2·tile_size × bpp`) is reverse-engineered from the engines' *private* pre-flight and had been copied across lib + scalability + pdf + flamegraph. Now **one** helper; the core-primitive ask is filed (**#47**).
- **Implementation quality**: solid — saturating arithmetic, child-isolated measurement, thorough docs. The helper is now transposition-resistant (`bpp: u32`, single `u64`) and `#[must_use]`.
- **Safety/reliability**: the width/canvas divergence (a latent `BudgetExceeded` re-trigger for centred plans) is closed. The `.run().unwrap()` report-wide-failure fragility is real but out of scope → **#46**.
- **Performance**: unchanged — the 2× budget still lands the engine at its minimum aligned strip (`compute_strip_height` returns `None`), so the true-streaming regime is preserved.
- **Maintainability**: one budget definition, honest docs, floor-vs-effective budget fields disambiguated.
- **Production-readiness**: green across build / fmt / clippy `-D warnings` / tests + runtime smoke. Ready pending the two follow-ups.

## Grouped findings + resolution

### MANDATORY (correctness) — applied
1. **Sized from `src.width()`, engine gates on `plan.canvas_width`** (Systems M, Safety L×2). Centred plans widen the canvas past the image, so the fix could re-trip `BudgetExceeded`. → every call site now passes `plan.canvas_width`; param renamed `canvas_width`; coupling guard test added.
2. **`write_libviprs_pyramid` passed the raw floor → PSNR spot-check silently skipped at ≥1024px** (Architect M). → budget now sized *inside* `write_libviprs_pyramid` via the shared helper; vips-free regression test added.
3. **Duplicate `streaming_budget_for` (lib pub vs scalability private)** (Architect M, Systems L, API M). → scalability's copy deleted, calls the shared saturating helper (algebraically identical `max(4 MB, w·ts·12)`; also removes its non-saturating overflow footgun).

### HIGH-VALUE OPTIONAL — applied
4. **`flamegraph.rs` still panics `BudgetExceeded` at 4096²** (Architect L). → sized the same way; smoke-verified both engines now run.
5. **Report header + snapshot field state a budget no size used** (Architect M, API L). → report prints the **floor + auto-scale policy**; `RunMetrics.memory_budget_bytes` documented as the **effective** per-run budget, `BenchmarkSnapshot.memory_budget_bytes` as the requested **floor**.
6. **2× doc claim contradicts the arithmetic** (Safety L, Systems L, Architect L, API L). → doc rewritten: the 2× is a fixed pre-flight margin (100% slack over the single-strip gate), **not** the full working-set estimate; the engine still runs at min strip via `compute_strip_height`'s `None`-fallback.
7. **Scalar-blindness + ergonomics** (API L + Suggestion). → `requested`→`floor`, `bpp`→`u32` (kills the floor/bpp two-`u64` swap + `as u64` noise), `#[must_use]`.
8. **Stale "mirroring" doc language** (Architect Suggestion). → rewritten as the single shared helper.

### DEFERRED → filed as issues
- **#46** — `bench_streaming`/`bench_mapreduce` still `.run().unwrap()`: any non-budget engine error panics the whole report + leaks a temp dir (Safety). Expert flagged out of scope.
- **#47** — adopt a core-exposed budget-admission primitive so the bench stops reverse-engineering the engines' private pre-flight (Architect M, API L). Needs a core-crate API addition.

### DECLINED (low value, one-liners)
- **`pub` → `pub(crate)`** (Systems Suggestion): `scalability` + `flamegraph` are separate bin crates and the integration test all call it, so `pub` is required — the expert's own "wire the reuse up" condition is now met.
- **Rename `streaming_budget_for` → `sized_budget_for`** (API Suggestion): it's the established shared name; kept as the streaming-family umbrella and said so in one doc line.
- **Add `requested_budget_bytes` / store per-size budgets in the snapshot** (API L, Architect M-alt): a serialized-schema change; the expert's stated minimum bar (documented meaning-change) is met instead.
- **`MIN_ALIGNED_STRIP_TILES`/`ACCUMULATOR_HEADROOM` consts** (API L): two literals in a 4-line fn, now fully explained in prose.

## Conflicting recommendations + resolutions
1. **Signature: plan-based (Systems, Safety, API) vs scalar.** → improved **scalar** `streaming_budget_for(floor, canvas_width, tile_size, bpp: u32)`. Plan-based would force throwaway plans at scalability's two plan-less display sites and couldn't express the saturation unit test; passing `plan.canvas_width` from the call sites achieves identical engine-gate parity with far less churn and keeps the helper usable from every binary.
2. **The 2× margin: drop to 1× for tightest measurement (Systems) vs size to `estimate_streaming_peak_memory` for the full working set (Safety, Architect).** → **keep 2× + honest docs**. Keeping 2× is *required* for the scalability consolidation to stay byte-for-byte identical (its floor formula is 2×); 1× would diverge from it, and the full estimate needs the plan-based form. All three agree the engine runs at min strip either way, so the measured regime is unchanged.

## Verification
`build` ✓ · `fmt --check` ✓ · `clippy --all-targets -D warnings` ✓ (+ `--features pdfium` lib) ✓ · `test` 61 passed / 0 failed / 2 ignored ✓ · doctests ✓ · runtime smoke: `flamegraph` runs at 4096² (was the live panic) and both budget-driven candidate pyramids build under the 1 MB floor ✓. Committed `8457579` `[skip ci]`; not pushed.

Issue #49, opened 2026-07-20
bench: schedule a non-gating run of the upstream libvips-pin check


spdrman:
Follow-up from the #35/#36 review (branch `bench/35-36-repro`).

The upstream-pin validator now exists as an on-demand check: the `check-libvips-pin` binary classifies a GitHub releases payload via `pin_check::classify_libvips_pin`, and `tools/check-libvips-pin.sh` fetches the feed and pipes it in (exit 0 up-to-date / 1 newer release or digest mismatch / 2 could-not-check). A live `#[ignore]`d test (`live_upstream_pin_is_current`) exercises the same path.

What is still missing is the *scheduling*: the only continuously-running check in CI is `recorded_pin_matches_the_captured_upstream_sample`, which compares the committed pin against the committed fixture — commit-time consistency, not live drift detection. Nothing runs the live check on a cadence, so an aging pin is only noticed when someone runs it by hand.

Proposed: wire a NON-gating scheduled job (e.g. a daily workflow, or a routine outside GitHub CI given this repo skips CI on PR commits) that runs `tools/check-libvips-pin.sh` (or `cargo test --test libvips_upstream_check -- --ignored`) and alerts on a non-zero exit, WITHOUT blocking merges. Deliberately advisory, matching the current design.

Refs #35, #36.

spdrman:
Resolved as won't-do per repo policy: this repo relies on local pre-commit gating and skips GitHub CI, so a scheduled/cron check is out of scope. The on-demand validator already exists (tools/check-libvips-pin.sh + the check-libvips-pin binary + the #[ignore]d live_upstream_pin_is_current test) and can be run locally at any time. No standing CI job by design.

PR #50, opened 2026-07-20
Dockerfile reproducibility: pin apt + validate the libvips pin against upstream (#35, #36)

### LINKED ISSUE #35: bench Dockerfile: snapshot-pin apt codec/toolchain packages (or digest-pin base images)
Follow-up to #24 (Dockerfile source-build review).

libvips is now pinned by exact version + SHA-256, and the base images are tag-pinned, but the builder-stage apt `-dev` packages libvips links against are **not** snapshot-pinned: `apt-get install libpng-dev libjpeg62-turbo-dev libtiff-dev libwebp-dev libglib2.0-dev libexpat1-dev build-essential` resolves against bookworm's live mirror. An intra-bookworm point release (e.g. a libpng security update) can therefore shift under a rebuild. libpng in particular sits on the DeepZoom PNG-tile encode hot path, so its point-release speed feeds directly into the measured libvips numbers.

Accepted for now and documented in the Dockerfile builder-stage comment: point releases hold ABI and rarely move the encode hot path materially, and the meson force-enable (`-Dpng/jpeg/tiff/webp=enabled`, landed in #24) hard-fails the build if a codec disappears entirely.

To fully close the reproducibility gap, pick one:
- Add a dated `snapshot.debian.org` apt source so the package set is fixed to a timestamp, or
- Digest-pin the `rust:1.89-bookworm` builder and `debian:bookworm-*-slim` bases (the pdfium stage is already date-tagged) so the apt snapshot they resolve against is fixed, and optionally record the linked libpng/glib versions into provenance so drift is visible.

### LINKED ISSUE #36: bench: periodic CI check validating the pinned libvips SHA/release against upstream
Follow-up to #24 (Dockerfile source-build review).

`tests/libvips_provenance.rs` validates only **internal string consistency**: Dockerfile `ARG LIBVIPS_VERSION` == `provenance::PINNED_LIBVIPS_VERSION`, Dockerfile contains `provenance::PINNED_LIBVIPS_SHA256`, and the `libvips-rs` binding major.minor == the pin. None of it verifies that the pinned SHA actually matches the real `vips-8.18.4.tar.xz`, that the release exists upstream, or that the image builds. A typo'd SHA, a nonexistent release, or a wrong URL passes `cargo test` green and only explodes at `docker build` — which the `[skip ci]` commit workflow does not exercise in CI.

Add a **periodic** (not per-PR) job that validates the pin against reality:
- minimum: fetch the upstream `vips-${VERSION}.tar.xz.sha256sum` companion (or the release asset) and diff it against `provenance::PINNED_LIBVIPS_SHA256`; or
- fuller: run a real `docker build` of the bench image on a schedule.

Note: the current `8.18.4` pin + SHA `2677bad6...ff037` were manually verified against the upstream release asset during the #24 review and match exactly — this issue is about making that check continuous rather than one-shot.

spdrman:
Resolves **#35** and **#36** (follow-ups from #24). #35: digest-pinned both base images (`debian@sha256`, `rust@sha256`) + snapshot.debian.org apt pinning on both stages, freezing libpng (the DeepZoom PNG hot path) and the codec/meson toolchain instead of floating them. #36: on-demand libvips-pin validator — `provenance::classify_libvips_pin` (single source of truth) + a `check-libvips-pin` binary + `tools/check-libvips-pin.sh` that shells out to it (no jq/bash reimplementation) + an `#[ignore]` live test. No CI cron added (repo policy).

TDD: RED tests → impl. 4-expert review applied (dual-classifier divergence fixed by the shared Rust binary). Clean rebase onto #38/charts. Local gate green.

Closes #35
Closes #36

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -64,6 +64,14 @@ sha2 = "0.11"
 name = "engine_comparison"
 harness = false
 
+# On-demand upstream-pin validator (libviprs-bench #35/#36). Reads a GitHub
+# releases JSON payload on stdin and classifies the recorded libvips pin via
+# `pin_check::classify_libvips_pin` — the single classifier the tests exercise.
+# `tools/check-libvips-pin.sh` curl-fetches the feed and pipes it here.
+[[bin]]
+name = "check-libvips-pin"
+path = "src/check_libvips_pin.rs"
+
 [[bin]]
 name = "flamegraph"
 path = "src/flamegraph.rs"
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,6 +28,13 @@
 #                     the live bookworm mirror (#35, closing the #33 scope gap)
 # The Rust and Debian base images are digest-pinned (`@sha256:`) on the FROM
 # lines below, so no layer floats on a mutable tag either (#35).
+#
+# Availability note (#35): pinning apt to snapshot.debian.org routes every
+# package fetch through a single archive host that is historically slow and can
+# be briefly rate-limited or degraded. That is the deliberate cost of a frozen
+# mirror; apt is set to retry and the libvips/PDFium tarball fetches use
+# `curl --retry`, so a transient snapshot blip should be re-run rather than
+# treated as a real failure.
 # ---------------------------------------------------------------------------
 ARG PDFIUM_RELEASE=pdfium-7881
 # Upstream libvips release compiled from source (issue #33). Kept in lockstep
@@ -52,6 +59,10 @@ FROM debian:bookworm-20250929-slim@sha256:7e490910eea2861b9664577a96b54ce68ea3e0
 # stage's `curl` resolves from a frozen mirror (#35). See the builder stage for
 # the full rationale; the deb822 source is rewritten wholesale (keeping the
 # stock archive keyring) so the pin is independent of the base image's layout.
+# NOTE: this deb822 snapshot block is duplicated verbatim in the builder stage
+# below (the canonical copy). The builder is `rust:*-bookworm` and cannot share
+# this Debian base image, so the two are kept intentionally in sync — edit both
+# together when either changes.
 ARG DEBIAN_SNAPSHOT
 RUN printf 'Types: deb\nURIs: http://snapshot.debian.org/archive/debian/%s\nSuites: bookworm bookworm-updates\nComponents: main\nSigned-By: /usr/share/keyrings/debian-archive-keyring.gpg\n\nTypes: deb\nURIs: http://snapshot.debian.org/archive/debian-security/%s\nSuites: bookworm-security\nComponents: main\nSigned-By: /usr/share/keyrings/debian-archive-keyring.gpg\n' \
         "${DEBIAN_SNAPSHOT}" "${DEBIAN_SNAPSHOT}" > /etc/apt/sources.list.d/debian.sources && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,13 +18,16 @@
 # Pinned inputs. A benchmark is only reproducible if every layer is fixed:
 # a floating base image, an unpinned libvips, or a `latest` PDFium would
 # silently change the numbers between runs (issue #153). Bump these
-# deliberately, never implicitly. (One documented exception — the apt codec
-# `-dev` packages libvips links against are not snapshot-pinned; that scope is
-# spelled out in the builder stage and tracked in #35.)
+# deliberately, never implicitly.
 #   PDFIUM_RELEASE  — libviprs-dep release tag (checksum-verified builder)
 #   LIBVIPS_VERSION — upstream libvips source release, built from tarball
 #   LIBVIPS_SHA256  — SHA-256 of that tarball, verified before it is built
-# The Rust and Debian base images are pinned by tag on the FROM lines below.
+#   DEBIAN_SNAPSHOT — snapshot.debian.org timestamp the apt toolchain resolves
+#                     against, so the codec `-dev` libraries (libpng et al. —
+#                     the DeepZoom PNG hot path) are fixed too, not floated on
+#                     the live bookworm mirror (#35, closing the #33 scope gap)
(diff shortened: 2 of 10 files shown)


spdrman:
## Consolidated review — #35/#36: Dockerfile reproducibility + upstream libvips-pin validation

Merged the four expert reviews (Principal Architect, Rust Systems, Safety/Reliability, API Design), grouped duplicates, resolved the conflicting recommendations, and applied every mandatory fix plus the high-value optional ones. Changes are in `d984ac8`; full suite green (build · `fmt --check` · `clippy --all-targets -D warnings` · 70 tests · binary smoke).

### Overall assessment
- **Architecture** — was the real weakness. The change added a *maintenance-time, remote-API* pin validator **and** a second copy of its logic in jq/bash, both wedged into the *per-run* `provenance` capture module. Resolved two ways: (1) the Rust `classify_libvips_pin` is now the **single** classifier, reached by a new `check-libvips-pin` binary that the shell script wraps; (2) the validator moved into a dedicated `pin_check` module, so `provenance` is capture-only again.
- **Implementation quality** — the core algorithm is sound (integrity-before-freshness, pre-release exclusion). Two genuine input-handling bugs fixed with regression tests.
- **Safety/reliability** — the jq path produced verdicts that diverged from the Rust classifier on realistic inputs; **empirically reproduced** three (null digest → false `MISMATCH`; mis-flagged higher pre-release → spurious `NEWER`; digest-absent pin → false `OK … SHA-256 matches`). Deleting the jq classifier removes the entire class.
- **Performance** — N/A to the benchmark hot path (this is maintenance tooling). The Dockerfile snapshot pin trades mirror speed for reproducibility; now documented as an availability dependency.
- **Maintainability** — net **−111 lines**; one classifier, one version-parse normalizer, docs made truthful.
- **Production readiness** — the check ships **on-demand** (advisory, matching repo policy of local gating + skip-CI on PR commits). Wiring it on a schedule is the one real gap, filed as **#49**.

### Conflicting recommendations & resolutions
1. **Single classifier — bin vs keep-bash.** Architect/Rust/API offered "add a Rust entrypoint and make the shell a thin wrapper"; the fallback was "keep bash as the sole validator, fix it in place." → **Resolved to the binary.** `tools/check-libvips-pin.sh` now only `curl`s the feed and pipes it to `check-libvips-pin` (which calls `classify_libvips_pin` over the compile-time pin constants). The jq/`extract_const` logic is deleted, so there is nothing left to drift.
2. **`Parse` error type.** Rust wanted `Parse(serde_json::Error)` + `source()`; API wanted `serde_json` kept out of the public type (`Box<dyn Error>`) **and** `Clone`/`PartialEq` added. → **Resolved to `Parse(serde_json::Error)` with `source()`** — it matches the crate's own `PdfBenchError` style and `serde_json` is already a public dependency. Declined `Clone`/`PartialEq` (a typed `serde_json::Error` is neither; tests already use `matches!`).
3. **`PinnedReleaseNotFound` naming.** Rust/API suggested rename or split. → **Kept the name** (crate is `publish = false`, no external match sites, and the doc already covers both cases) but the binary prints an explicit *"COULD NOT VERIFY … not found in the feed window"* so operators get the clarity without public-API churn.

### Mandatory — applied
- **[ALL 4 experts] Dual classifier divergence.** Collapsed to one `classify_libvips_pin`; shell delegates to the `check-libvips-pin` binary; parallel jq deleted; docs corrected ("invokes", not "reimplements"). Guard test rewritten to assert delegation (and that no jq `sort_by` ranking returns).
- **[Rust, Safety] Tarball name built from the raw arg.** `classify_libvips_pin` now rebuilds `vips-<major.minor.patch>.tar.xz` from the *parsed* tuple, so a `v8.18.4`/`vips-8.18.4` pin validates instead of falsely reading as `PinnedReleaseNotFound`. Regression test added.
- **[Rust, Safety] Digest lookup ignored draft/prerelease.** Added `!r.draft && !r.prerelease` to the digest filter so a draft can't drive a false `Sha256Mismatch`. Regression test added.
- **[Architect Medium] Mismatched module concern.** Extracted `classify_libvips_pin` + status/error enums + GitHub DTOs + `parse_libvips_version` into a new `pin_check` module; `provenance` module doc is accurate again.

### High-value optional — applied
- **[Rust, API] `Parse` severs the error chain.** Now wraps `serde_json::Error` and returns it from `Error::source()`.
- **[Architect, Rust, API] Two near-duplicate version parsers.** Unified behind `strip_libvips_prefixes` (strips `vips-` then optional `v`); the 2- and 3-component parsers now accept identical prefixes. Pins `parse_libvips_major_minor("v8.18")`.
- **[Safety, Architect] `base_image_date` read from a comment.** Now reads the comment-stripped instructions, so header prose can't stand in for the real `FROM` tag; documented that it keys off the tag (a digest's build date isn't knowable offline).
- **[Safety] Fixture test mistaken for drift detection.** Doc clarified: it is commit-time consistency; live drift is the `#[ignore]`d test + the script.
- **[Architect Low] Dockerfile.** Header now notes the `snapshot.debian.org` availability dependency; the duplicated deb822 block is marked as intentionally kept in sync with the builder stage.

### Optional — declined (with reason)
- **`#[non_exhaustive]` on the public enums** — crate is `publish = false`; no external `match` sites, so it adds ceremony with no stability benefit.
- **Rename `parse_libvips_version` / split `PinnedReleaseNotFound`** — public-API churn across call sites for a cosmetic gain; docs + the binary's operator message already disambiguate.
- **`LibvipsPin` newtype / struct-of-independent-signals / drop echoed `pinned` / distinct unparseable-pin error** — the API expert notes these are acceptable as-is for the advisory CLI purpose; deferring until genuine programmatic use exists.
- **Dockerfile full factor-out (shared base / COPY script) + retry-backoff + BuildKit cache mount** — Low-rated and not Docker-buildable in this environment; the sync comment covers the drift risk. CI cache/throughput can be picked up alongside #49 if desired.

### Follow-up filed
- **#49** — schedule a non-gating run of `check-libvips-pin` (the check exists on-demand; nothing runs it on a cadence yet).

PR #51, opened 2026-07-20
Budget hardening: fail-soft engine runs + budget-coupling reconcile (#46, #47)

### LINKED ISSUE #46: bench: bench_streaming/bench_mapreduce still .run().unwrap() — a non-budget engine error panics the whole report and leaks a temp dir

### LINKED ISSUE #47: bench: streaming budget admission is reverse-engineered from the core engine's private pre-flight — adopt a core-exposed primitive

spdrman:
Resolves **#46** (fixed) and **#47** (documented won't-fix). #46: bench_streaming/bench_mapreduce (and monolithic) now return `Result` instead of `.run().unwrap()`, with a self-reclaiming `TempOutDir` RAII guard (unwind-safe, no temp-dir leak on the error path); callers unified behind `skip_on_engine_fault`/`warn_engine_skip` so a failed cell degrades to skipped+warned rather than aborting the report/sweep. #47: evaluated core — no suitable public primitive exists (only the forward/full-working-set estimates, not the inverse); kept the tested formula and added a reality-pin test (`engine_admission` drives BOTH streaming+mapreduce, asserts rejection at min-1 is specifically `BudgetExceeded`) that goes red if core's admission rule drifts. Documented rationale on the helper (strawman corrected); `budget`→`budget_floor` rename for clarity.

TDD RED→GREEN. 4-expert review applied. **Convergence mode: no new issues filed.** Local gate green.

Closes #46
Closes #47

### DIFF
--- a/src/flamegraph.rs
+++ b/src/flamegraph.rs
@@ -118,7 +118,7 @@ fn generate_flamegraph(stacks: &[String], output_path: &Path, title: &str) {
     flamegraph::from_reader(&mut opts, reader.as_bytes(), writer).expect("generate flamegraph");
 }
 
-fn main() {
+fn main() -> Result<(), Box<dyn std::error::Error>> {
     let report_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("report");
     fs::create_dir_all(&report_dir).unwrap();
 
@@ -129,8 +129,9 @@ fn main() {
     // Size the streaming/mapreduce budget to admit the worst-case tile-aligned
     // strip at this 4096-wide canvas: a flat 1 MB trips `BudgetPolicy::Error`'s
     // `BudgetExceeded` up front (worst-case strip = 4096·512·3 = 6.29 MB) and
-    // `.run().unwrap()` would panic — the exact failure mode issue #38 fixes.
-    // RGB8 gradient, so bpp = 3; shared with the report/scalability paths.
+    // the engine `.run()?` below would surface it as an error — the exact
+    // failure mode issue #38 sizes around. RGB8 gradient, so bpp = 3; shared
+    // with the report/scalability paths.
     let budget = streaming_budget_for(1_000_000, plan.canvas_width, TILE_SIZE, 3);
 
     println!("Image: {WIDTH}x{HEIGHT}, tile_size={TILE_SIZE}");
@@ -151,8 +152,7 @@ fn main() {
             .with_engine(EngineKind::Monolithic)
             .with_config(EngineConfig::default())
             .with_observer_arc(observer.clone())
-            .run()
-            .unwrap();
+            .run()?;
         let elapsed = start.elapsed();
 
         println!(
@@ -185,8 +185,7 @@ fn main() {
             .with_memory_budget(budget)
             .with_budget_policy(BudgetPolicy::Error)
             .with_observer_arc(observer.clone())
-            .run()
-            .unwrap();
+            .run()?;
         let elapsed = start.elapsed();
 
         let strip_count = observer
@@ -226,8 +225,7 @@ fn main() {
             .with_memory_budget(budget)
             .with_budget_policy(BudgetPolicy::Error)
             .with_observer_arc(observer.clone())
-            .run()
-            .unwrap();
+            .run()?;
         let elapsed = start.elapsed();
 
         let events = observer.events();
@@ -261,4 +259,6 @@ fn main() {
 
     println!();
     println!("Flame graphs written to {}", report_dir.display());
+
+    Ok(())
 }
--- a/src/harness.rs
+++ b/src/harness.rs
@@ -78,28 +78,13 @@ impl Engine {
     }
 }
 
-/// Degrade a fail-soft streaming / mapreduce cell to a skipped row: on a genuine
-/// engine fault (the strict `BudgetPolicy::Error` these engines run under, a
-/// sink IO error, the plan/source dimension guard, …) record no metrics and
-/// warn, so the parent driver drops just that cell instead of the child
-/// panicking and taking the whole engine series down with it (issue #46).
-fn warn_on_skip(label: &str, res: Result<RunMetrics, libviprs::EngineError>) -> Option<RunMetrics> {
-    match res {
-        Ok(m) => Some(m),
-        Err(e) => {
-            eprintln!("cell {label} failed, skipping (issue #46): {e}");
-            None
-        }
-    }
-}
-
 /// Run exactly one cell in *this* process and return its metrics.
 ///
 /// This is the child body behind `--single`. Each engine is exercised once
 /// (single shot); the parent handles warm-up, repetition, and aggregation.
-/// A streaming / mapreduce engine fault degrades to `None` (a skipped cell)
-/// rather than panicking, so one unmeasurable cell never aborts the report
-/// (issue #46).
+/// Any engine fault degrades to `None` (a skipped cell) via the shared
+/// [`skip_on_engine_fault`](crate::skip_on_engine_fault) rather than panicking,
+/// so one unmeasurable cell never aborts the report (issue #46).
 pub fn run_single_cell(spec: CellSpec) -> Option<RunMetrics> {
     use libviprs::{Layout, PyramidPlanner};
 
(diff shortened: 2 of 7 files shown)


spdrman:
## 4-expert review of #46/#47 — consolidated & applied

All four experts converged on one dominant defect and a cluster of smaller ones. Applied in `bf31694` (branch `bench/budget-converge`). No new issues filed (convergence sweep); every non-fix is marked WON'T-FIX/ACCEPTED inline below.

### Overall
The PR does what #46/#47 promise for **2 of 3** raster engines and pins #47 behaviourally — solid work. The gaps are (a) the monolithic path was left on the exact `.run().unwrap()` + leak anti-pattern #46 removes, (b) the fail-soft "skip" policy was copied 5 ways with drifting messages, (c) the #47 rationale leaned on a strawman, and (d) the "budget" param is really a floor. All addressed.

### Mandatory (unanimous / Medium+) — APPLIED
1. **Monolithic asymmetry** *(API High; Architect/Systems/Safety Medium — unanimous)*. `bench_monolithic` and `scalability::run_monolithic` still did `.run().unwrap()` with cleanup *after* the unwrap → a sink-IO / `PlanSourceMismatch` fault both panicked and leaked `$TMPDIR`. Both now return `Result`/`Option`, fail-soft at every call site (comparison_suite, `run_single_cell`, the scalability sweep), reclaiming the dir first. The invariant "one bad cell never aborts the report" now holds for all three engines. The experts offered "harden **or** document-exempt + file issue" — convergence rule forbids the issue, so hardened symmetrically (the cheaper long-term choice anyway).
2. **Fail-soft helper duplicated 5× w/ drifting messages** *(Architect/API Medium; Systems Low)*. Hoisted `skip_on_engine_fault` (+ a shared `warn_engine_skip` logger) to the lib root; harness and comparison_suite route through it; scalability's `Option`-returning runners share the one logger. Five wordings → one: `"<context> failed, skipping: <err>"`.
3. **`(issue #46)` in operator stderr** *(Architect Low)*. Dropped from every printed line; issue refs kept in code comments.
4. **#47 docstring strawman** *(Architect Medium; Systems Suggestion)*. The old rationale claimed adopting a core primitive "would change the sized value (full-working-set estimate)" — false: the bench re-derives `worst_case_strip_bytes` exactly (core `streaming.rs:928` / `streaming_mapreduce.rs:505`), which a core accessor would return unchanged. Rewritten to the honest reasons: core exposes no public accessor (a core-scope change), and its natural `(&PyramidPlan, format)` signature wouldn't fit this deliberately-raw plan-less helper.
5. **"budget" param is a floor** *(Architect/API Medium — the core of #47)*. Renamed `memory_budget_bytes`→`budget_floor_bytes` on `bench_streaming`/`bench_mapreduce`/`bench_streaming_pdf`/`write_libviprs_pyramid` (+ `comparison_suite`'s param), documenting that the effective budget is derived and reported via `RunMetrics::memory_budget_bytes`. Chose the rename over "invert so callers size" (both experts' shared low-risk recommendation; inverting would break the #38 tests' semantics).

### High-value optional — APPLIED
6. **RAII temp-dir guard** *(Systems/Safety Low)*. `fs_sink_dir` now returns a self-reclaiming `TempOutDir`; the 4 hand-placed `remove_dir_all`-before-`?` lines are gone. Leak-freedom is now **unwind-safe** (the manual version only covered the returned-`Err` channel) and enforced by the type across all four bench paths.
7. **#47 pin strengthened** *(Architect Low; Safety/Architect Suggestions)*. The pin drove streaming only and asserted `is_err`. Now drives **both** engines (core keeps two independent gate copies that can drift) and asserts the min−1 rejection is specifically `BudgetExceeded` (was passing-for-the-wrong-reason risk). Dropped the tautological `sized >= min_strip`; reframed `assert_eq!(sized, min_strip*2)` as documenting the margin, not a core pin.
8. **#46 test only proved non-panic** *(Systems Low)*. Now binds the returned `Result` and asserts `Err(PlanSourceMismatch)` — pins that the fault is *surfaced as a skip signal*, not swallowed into a spurious `Ok` (garbage report row).

### Documented / partial
- **Ownership split (bench_* size internally vs scalability run_* take a final budget)** *(Architect Medium)* — documented both contracts; did not unify to a single sizing site (larger refactor, and the #38 tests pin the internal-sizing semantics).
- **Raster collapses all `EngineError` variants vs `PdfBenchError` discipline** *(API Medium)* — took the expert's "document the decision at the return type" branch rather than a new enum: in the real gradient drivers the plan is always built from the source's own dims (`PlanSourceMismatch` unreachable) and the budget is pre-sized (`BudgetExceeded` can't trip), leaving only environmental sink-IO faults that are legitimately skippable — so, unlike the PDF path, there's no distinct always-skip case to separate.
- **`fs_sink_dir` unwrap/uniquifier** *(Safety Suggestion, "no action required today")* — added the accepted-environmental-panic note; no uniquifier (labels are unique per cell, cells run serially).

### WON'T-FIX (rationale; no issues per convergence rule)
- **Expose/dedupe a core `min_streaming_budget` primitive** *(Architect Medium; Systems Suggestion)* — out of scope for a bench-only PR, and the raw plan-less helper couldn't adopt a `(&PyramidPlan, format)` core fn anyway. Mitigated instead by #4 (honest docs) + #7 (both core gate copies now pinned behaviourally, so drift in either goes red). Core left untouched.
- **Surface skips in the durable report/scalability artifact + non-zero exit** *(Safety Low)* — beyond this hardening PR's scope (touches the serialized schema); the now-unified stderr warnings surface every skip and benchmarks run under operator supervision.
- **`streaming_budget_for` struct-args / `From<PixelFormat>` for bpp** *(API Suggestion, self-flagged low-priority)* — arg order is documented and the behavioural pin guards the arithmetic; the raw positional signature is deliberately plan-less for the scalability chart annotations.
- **`create_snapshot` `memory_budget_bytes`** — populates the serialized `BenchmarkSnapshot` schema field (report-wide configured budget, a distinct concept); renaming is a schema change, out of scope.

### Conflicts & resolution
The only real tension was #47's core-primitive ask (Architect/Systems) vs the convergence rule (no new issues) and the helper's deliberately-raw signature. Resolved by *not* adopting/exposing a core primitive and instead making the docs honest + pinning **both** engine gates behaviourally — which is what actually protects against core drift.

### Verification (all green, `RUSTFLAGS='-L /opt/homebrew/lib'`)
`cargo fmt --check` ✓ · `clippy --features libvips --all-targets -D warnings` ✓ · `build`/`check --features libvips,pdfium` ✓ · `test --features libvips` (19 lib + all integration incl. the 3 hardened budget tests) ✓ · `--doc` ✓.

PR #52, opened 2026-07-20
Finish JS chart migration: grouped-bar charts, drop plotters, zoom, golden drift test (#42, #43, #44)

### LINKED ISSUE #42: charts: finish the JS migration — port grouped-bar chart_*.svg to render.mjs and drop plotters

### LINKED ISSUE #43: charts: restore large-image x-axis zoom in the JS scalability renderer (supersedes --crop)

### LINKED ISSUE #44: charts: golden round-trip test to catch Rust↔JS producer/consumer field-shape drift

spdrman:
Resolves **#42**, **#43**, **#44**. #42: ported the grouped-bar chart_*.svg renderers (all 6, incl. tracked-memory + 95% CI whiskers) to `tools/charts/chart.mjs`, render.mjs emits them from benchmark_results.json, removed `generate_charts` + the plotters block from lib.rs/report.rs, and dropped plotters+criterion from `[dependencies]` — **plotters is now fully out of the normal build graph** (`cargo tree -i plotters -e normal` empty). #43: `--zoom <minMP>` / `xMin` additively writes large-image `scalability_*_zoom.svg`; log-log full-range stays default. #44: committed golden fixtures + a two-sided drift guard (JS `resultsShapeOk` + Rust `chart_shape_drift.rs` vs the real serializers) catching Rust↔JS field-shape drift.

TDD RED→GREEN. 4-expert review applied (restored the 6th chart + CI whiskers + drift-guard reachability). **Convergence mode: no new issues filed.** Clean rebase onto budget-conv. Local gate + 66 JS tests green.

Closes #42
Closes #43
Closes #44


spdrman:
# Consolidated review — Finish JS chart migration (#42/#43/#44)

Four expert passes (Architecture, Rust Systems, Safety/Reliability, API Design), de-duplicated and grouped below. Every mandatory item plus all high-value optional ones are applied on `bench/charts-converge` (HEAD `8f2f7aa`). Full matrix green: `cargo fmt` / `clippy --features libvips --all-targets -D warnings` / `test --features libvips` (69 lib+integration, 1 doc) + `node --test tools/charts/*.test.mjs` (66).

## Blocking / high (fixed)

**1. `drop plotters` was not actually met — the code asserted a falsehood** (Rust Systems: High; Safety: Low). `plotters` was still in the NORMAL build graph via `criterion`'s `html_reports`: `criterion` was a regular `[dependencies]` entry *and* a `[dev-dependencies]` one, yet nothing under `src/` uses it (only the dev-scope `benches/engine_comparison.rs`). So every `cargo build --bin report` still compiled plotters, and the "plotters is gone" claims in `report.rs`/README were untrue; the src-only grep guard couldn't see the Cargo.toml vector. **Fix:** removed the normal `criterion` dep (kept the identical dev-dep). `cargo tree -i plotters -e normal` is now empty (plotters survives only in the criterion micro-benchmark's dev graph), which makes the existing claims *accurate* rather than merely softened. Added `plotters_stays_out_of_the_normal_build_graph`, which parses `[dependencies]` and fails if `plotters`/`criterion` reappears there.

**2. One of six comparison charts silently dropped** (Architecture: High; Rust Systems / API Design: Medium). The retired Rust `generate_charts` emitted six grouped-bar charts; the port emitted five — `chart_tracked_memory.svg` (Engine-Tracked Working Set) was gone, though `tracked_memory_mb()` is still first-class in the text table and cross-version report, and the golden test cemented the five-set. *Tension:* API Design offered "document an intentional deprecation" as an alternative. **Resolved toward parity restoration** (`runTrackedMemoryMb` extractor + `renderTrackedMemoryBars` wrapper + `chart_tracked_memory.svg`), since three passes flagged it as a regression and the metric remains first-class elsewhere — the no-capability-loss path.

## Medium (fixed)

**3. 95% CI whiskers dropped** (all three systems passes). Wall-time/peak-RSS charts lost the CI whiskers the Rust charts drew from `RunStats.wall_ms_ci95/rss_mb_ci95` (data present in the JSON but unread). Widened the row contract to `{config, engine, value, error?}`; a `#333` whisker renders when `error>0`, fed from ci95 for those two metrics only (ratios carry none, as in Rust). Titles restored to "whiskers = 95% CI".

**4. Runtime shape-drift guard was unreachable** (Rust Systems, Safety). `main()` returned as soon as `written.length>0`, so `resultsShapeOk` never ran on real input — a partial rename (e.g. `peak_rss_bytes`) still rendered wall-time charts, exited 0, and drew the rest as all-zero bars, contradicting the "FAILS LOUD" docstring. Moved the drift gate ahead of rendering: it runs on every present input and exits non-zero writing nothing on drift. Added two subprocess tests (drift → exit 1, no charts; golden → exit 0, six charts).

**5. Fixed 880px canvas** (Architecture; Rust Systems/Safety: Low; API Design). Non-responsive width collided labels at the default 8-config sweep and produced negative-width (invalid) `<rect>`s past ~48 configs. Content-sized the canvas from config×engine count (mirroring the old `chart_w = 160 + n*(max_bars*35+50)`), with a clamped bar width and a label-suppression floor. Added a 60-config test asserting all rect widths stay positive and the viewBox grows past 880.

**6. Per-metric wrappers broke the `(data, opts)` convention** (API Design). They took `rows` only, so `render.mjs` could never set width/title. Now `(rows, opts = {})` merging fixed title/unit with forwarded opts.

**7. No value oracle for the derived extractors** (Architecture). The drift guard checks field shape only; throughput/efficiency/resource-cost arithmetic had no oracle. Extended the golden round-trip to assert hand-computed labels (throughput 1000, efficiency 5, resource-cost 0.2, tracked 16MB) and cross-referenced the Rust `RunMetrics` methods in the extractor comment.

## Low / suggestions

- **O(N²) config rescan** (Arch/Rust/Safety) — single-pass `Map<config, runs[]>`. *fixed*
- **Absent vs measured-zero cell** (Rust Systems) — `byKey.has` distinguishes them; an absent cell keeps its aligned zero-height slot but draws no misleading "0" label (a measured zero still labels honestly). *fixed*
- **`resultsShapeOk` accepted a non-array top level** (API Design) — all three probes now treat a present-but-non-array payload as drift; `null`/`[]` remain "no signal". *fixed*
- **Stale wrapper enumerations** (API Design) — `renderEfficiencyBars` (+ the new tracked-memory wrapper) added to both lists. *fixed*
- **`walk_rs` symlink cycle** (Safety) — recurse via `file_type().is_symlink()`, never following links. *fixed*
- **Throughput unit unstated** (API Design) — title now "Raw Throughput — Tiles/s". *fixed*
- **`zoom`/`xMin` silent rename** (API Design) — documented at the `renderAll` boundary. *fixed*
- **Duplicated default-path joins** (Rust Systems) — hoisted into a shared `resolveInputPaths`. *fixed*
- **Floating bars / no baseline** (Rust Systems) — added a faint baseline at the bar footing. A full y-axis + gridlines was intentionally NOT added: the port deliberately follows the flat causl-bench aesthetic, and per-bar labels + baseline give the magnitude reference. *partial (accepted)*
- **peak_rss `0=UNKNOWN` sentinel differs from the history path** (API Design) — kept the comparison path's `0` return to MIRROR the Rust `peak_rss_mb()` so chart numbers match the text table; documented that a fresh `benchmark_results.json` (the only input here) always carries RSS, so the sentinel cannot arise. *documented, no behavior change*

## Assessment
A solid migration whose two material gaps were (1) plotters never actually leaving the normal build graph and (2) the silently-dropped sixth chart — both now closed with parity restored and locked by tests, alongside a genuine correctness fix to the runtime drift guard. Per the convergence goal, no new issues were filed; the two items left unchanged are documented in-code with rationale.

Issue #33, opened 2026-07-20
Dockerfile: build + record a pinned modern libvips


spdrman:
Sub-issue of #24. The container benches against Debian's frozen libvips 8.14.1 with mismatched 8.18 bindings. Build libvips from a pinned recent source tarball (8.16/8.17.x) in the Dockerfile, align the libvips-rs bindings, and make LIBVIPS_VERSION a recorded axis (into provenance). Test: provenance records the measured libvips version; a check asserts it matches the pinned tag.

PR #53, opened 2026-07-20
Residual sweep: metric labels, tables, thermal/load provenance, flamegraph weighting, docs (#25)

### LINKED ISSUE #25: Sweep the remaining ~50 lower-severity benchmark review findings
The fairness rework addressed the 15 criticals/highs from the 4-expert review. ~50 residual medium/low items (chart styling, metric labeling e.g. pixels/s as primary throughput, table formatting, cold-vs-warm series, thermal/load recording, flamegraph time-weighting, Docker digest pinning, doc polish) were not individually triaged — sweep the full report and file/fix the residuals.

spdrman:
Resolves **#25** — the final residual sweep of the lower-severity benchmark-review items. (1) Metric labeling: units + higher/lower-is-better on every column (report tables, comparison_table.txt, cross_version, JS charts); primary throughput disambiguated (T = pyramid tiles, never pixels). (2) Table formatting: thousands separators + widened columns. (3) Cold/warm: warm-median disclosure + iters/warmup policy in the committed artifact. (4) Thermal/load: `Provenance.load_average` + `thermal_throttle_count` (serde-default, excluded from fingerprint, surfaced as contention/throttle warnings). (5) Flamegraph: corrected to true time-weighting (µs by inter-tile gap; count-as-weight markers removed; per-tile claim scoped to the serial engine). (6) Docs: fixed stale README/module/run-bench.sh references (pdf_engine_bench→pdfium_strip_source_bench, plotters/crop/replot).

TDD RED→GREEN. 4-expert review applied. **Convergence mode: no new issues filed.** Local gate + 68 JS tests green.

Closes #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The harness produces:
 | `benches/engine_comparison.rs` | Criterion micro-benchmarks for monolithic / streaming / MapReduce, plus a head-to-head group across image sizes. |
 | `src/scalability.rs` (`scalability` bin) | Scalability sweep from 0.2 MP to 47 MP, comparing all four engines on a 1.42:1 aspect ratio matching `43551_California_South.pdf`. |
 | `src/report.rs` (`report` bin) | Full comparison matrix across image sizes and concurrency levels, with versioned history. |
-| `src/flamegraph.rs` (`flamegraph` bin) | Event-based flame graphs for all three libviprs engines on a 4096x4096 image. |
+| `src/flamegraph.rs` (`flamegraph` bin) | Time-weighted flame graphs (frame width = per-tile µs) for all three libviprs engines on a 4096x4096 image. |
 
 ## Running benchmarks
 
@@ -143,22 +143,27 @@ Four benchmark groups, each parameterised on image size:
   [`--memory-budget`](https://libviprs.org/cli/#flag-memory-budget) /
   [`--concurrency`](https://libviprs.org/cli/#flag-concurrency) tradeoff.
 
-### `pdf_engine_bench` (`src/pdf_engine_bench.rs`)
+### `pdfium_strip_source_bench` (`src/pdfium_strip_source_bench.rs`, `pdfium` feature)
 
-Renders one PDF page through one engine and writes a JSON report. The CLI
-flags it accepts mirror the libviprs CLI:
+Compares the two `PdfiumStripSource` constructors — cached (`new`) versus
+streaming (`new_streaming`) — over a `(dpi, strip_count)` sweep on a PDF,
+emitting one newline-delimited JSON record per `(mode, dpi, strips)` triple. The
+numbers back the "vector-heavy PDFs scale ~linearly with N, raster-heavy
+approach 1×" doc comment on `PdfiumStripSource::new_streaming` with data instead
+of assertion. Gated behind the `pdfium` feature.
 
-| `pdf_engine_bench` flag | CLI equivalent                                              |
-|-------------------------|-------------------------------------------------------------|
-| `--engine mono\|stream` | engine selection (see CLI flag picker)                      |
-| `--budget-bytes`        | [`--memory-budget`](https://libviprs.org/cli/#flag-memory-budget) |
-| `--tile-size`           | [pyramid layout](https://libviprs.org/cli/#pyramid)         |
-| `--dpi`, `--page`, `--layout` | matched 1:1 in the CLI                                |
+| `pdfium_strip_source_bench` flag | Meaning                                                                  |
+|----------------------------------|--------------------------------------------------------------------------|
+| `--pdf <path>`                   | PDF to bench (default: the committed `fixtures/cc_licenses_mapping.pdf`)  |
+| `--page <N>`                     | 1-based page index                                                       |
+| `--dpis 72,150,300`              | comma-separated render DPIs to sweep                                      |
+| `--strip-counts 4,16,64`         | comma-separated strip counts to sweep                                    |
+| `--output <file.jsonl>`          | write JSONL here instead of stdout                                       |
 
-Streaming runs additionally honour an internal strip-buffer sizing knob
-analogous to [`--buffer-size`](https://libviprs.org/cli/#flag-buffer-size);
-the bench fixes it via `compute_strip_height` against the budget so the
-two engines see byte-identical input dimensions.
+```bash
+cargo run --release --features pdfium --bin pdfium_strip_source_bench -- \
+    --pdf /path/to/blueprint.pdf --dpis 72,150,300 --strip-counts 4,16,64
+```
 
 ## Docker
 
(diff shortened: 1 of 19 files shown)


spdrman:
## Consolidated 4-expert review — libviprs-bench #25 (residual sweep)

Four expert passes (Principal Architect, Rust Systems, Safety/Reliability, API/Library Design) on `bench/25-sweep`. Findings converge on **three substantive themes** plus a cluster of formatting/naming polish. No hard conflicts between reviewers — where they differed it was only on *how far* to take a fix. All mandatory items are applied; every declined item is marked WON'T-FIX/ACCEPTED inline (no new issues filed, per the convergence rule).

### Priority 1 — Provenance captured at the wrong moment, and twice (API **High** + Rust Systems **Medium**)
The durable snapshot re-called `Provenance::capture()` *inside* `create_snapshot_for` (post-workload), while the console banner + `comparison_table.txt` used a `prov` captured *pre-workload*. With #25's new dynamic load/thermal axes the two committed artifacts of one run could disagree on host load, and the persisted value recorded the tail of the benchmark's **own** load curve instead of the ambient contention its doc promises.
**Fix (mandatory, applied):** `create_snapshot` / `create_snapshot_for` now take an injected `Provenance`; `report` passes the single start-of-run `prov`; `version_matrix::append_version_snapshot` captures live in its own process and injects it. One capture per run → banner, warnings, `.txt`, and history all agree. This also structurally removes the post-run "load ≥ ncpu self-trigger" the reviewers flagged.
*Declined:* the "sample at start **and** end, store the pair/max" enhancement — needs new serialized fields + a two-sample model; single-capture-at-start already records the honest ambient pre-run condition and makes persisted == printed.

### Priority 2 — Flame-graph per-tile time-weighting is only valid for a serial producer (Architect + Rust Systems + Safety, all **Medium** — unanimous)
`flame.rs` asserted unconditionally that "the wall-clock gap between one tile and the next is that tile's production time," and all three SVGs were titled "time-weighted (width = µs)". That holds for the **monolithic** engine (serial row-major emission) but is false for the strip-based **streaming/MapReduce** engines: each renders a whole strip then emits its tiles back-to-back (first tile absorbs the strip render, the rest drain in ~0 gaps), and under `tile_concurrency > 0` `TileCompleted` is emitted in channel-arrival order the core itself documents as arbitrary. `events_to_folded_stacks` is public, so any caller can feed a concurrent stream.
**Fix (mandatory, applied):** scoped the claim honestly — module + fn docs now state per-tile widths are faithful only for the serial monolithic engine; for the strip-based engines only the aggregate (level/root width = total tiling time) is a true wall-time measure. The three SVG titles now differentiate ("per-tile width = µs; serial" vs "read at level/root — per-tile widths are emission cadence"); README updated. Added a `batch_burst_*` test pinning both halves: the boundary tile absorbs the batch latency (per-tile reading lies) **and** the weights sum to the wall-clock span (aggregate stays true).
*Declined:* driving MapReduce at real concurrency / a `worker_id`-based flat-weighting fallback — in-tree `worker_id` is always `None` so a fallback can't trigger, and raising concurrency would make per-tile widths meaningless anyway. The documentation fix is the low-effort honest option all three reviewers endorsed.

### Priority 3 — Duplicated, already-drifted measurement warnings (Architect **Low** + API **Medium**)
The contended-host + thermal (+ #33 oracle-mismatch) stderr blocks were copy-pasted across `report` and `scalability` and had **already** diverged ("this run" vs "this scalability run", "host load" vs "Host load", differing thermal tails).
**Fix (applied):** added `Provenance::measurement_condition_warnings(&self) -> Vec<String>` as the single source of wording; both binaries now `for w in prov.measurement_condition_warnings() { eprintln!("{w}") }`. Added a unit test pinning the contract. (`cross_version`'s per-snapshot mismatch warning is intentionally left separate — different subject: a historical snapshot named by version+timestamp, not "this run".)

### Cluster — labeling / formatting / naming polish
| Finding | Reviewers | Sev | Status |
|---|---|---|---|
| Tiles/T-s `{:>8}` columns overflow once thousands separators push a value ≥ 9 chars ("1,048,576") | **all 4** | Suggestion | **Applied** — widened Tiles/T-s to 11 in `comparison_table` + savings summary, eff column to 14 in verdict, Tiles to 11 in scalability summary; separators fixed to match |
| Thermal reads only `cpu0`, and warning asserts "during THIS run" (counter is cumulative-since-boot) | Architect / RustSys / API | Sugg/Low/Med | **Applied** — now max over all `cpu*` `core_throttle_count` + `package_throttle_count`; wording softened to "throttled at some point since boot", no run-scoped causation. Start/end delta declined (schema churn) |
| `thousands` is an ambiguous public name | API | Low | **Applied** — renamed `format_thousands` |
| `LoadAverage` fields `one/five/fifteen` carry no unit | API | Low | **Applied** — `one_min/five_min/fifteen_min` (golden fixture + tests updated; JS charts don't read the subfields) |
| Shared legend asserts "warm medians" — false for future single-run callers; `report.txt` states it twice | Architect | Low | **Applied** — split into `COMPARISON_TABLE_LEGEND` (units) + gated `COMPARISON_TABLE_WARM_NOTE` (appended only when every row has stats); trimmed the `.txt` header duplicate |
| Empty-stack run panics (`inferno::from_reader` on empty input) — robustness regression from dropping marker frames | Safety | Low | **Applied** — `generate_flamegraph` guards `stacks.is_empty()`, prints a notice, skips |
| `getloadavg` unsafe block lacks a `// SAFETY:` note | RustSys | Suggestion | **Applied** |
| Notation drift: `T/s` vs `Tiles/s`, `T/s/MB` vs `T/s/RSS-MB` | API | Low | **Partial** — unified the RSS-MB basis everywhere (verdict `T/s/MB`→`T/s/RSS-MB`, restated the `T = tiles` key); kept `T/s` (Rust tables, with key) vs spelled-out `Tiles/s` (standalone chart SVGs) as an intentional per-surface convention |
| `tile_weight_micros(prev, cur)` transposition footgun | API | Suggestion | **Applied** — one-line arg-order doc note |
| Contention: single lagging 1-min sample can't prove run-scoped contention | RustSys / API | Low/Med | **Applied** — single-capture-at-start (Priority 1) removes the self-trigger; `host_looked_contended` doc + warning reworded as a pre-run ambient sample. Threshold kept at `>= ncpu` (accepted: pre-run ambient load ≥ ncpu genuinely indicates external contention) |
| `Provenance` mixes stable identity with transient load/thermal; nest a `RunConditions`? Eq dropped | API | Suggestion | **Accepted / won't-fix** — flat `#[serde(default)]` fields are forward-compatible and documented as per-run conditions; nesting would churn the persisted schema for organizational purity only. Eq removal already documented intentional |

### Overall assessment
A strong, well-converged review. The one **High** (persisted-provenance inconsistency) is a genuine correctness bug for the committed history artifact and is fixed at the root; the unanimous flame-graph **Medium** is an honesty/labeling defect resolved by scoping the claim and differentiating the titles (with a regression test); everything else is polish that measurably improves legibility and de-duplicates drift-prone wording. No behavioural regressions — the engines, budgets, and JSON shapes are untouched; the only serialized-schema change (`LoadAverage` keys) rides on a field introduced in this same unmerged PR, so there is no real history to migrate. **Verification:** `fmt --check`, `clippy --features libvips -D warnings`, `cargo test --features libvips` (lib+integration+doc, 23 binaries), and `node --test` (68 JS) all green; the `flamegraph` binary was driven end-to-end and emits the corrected titles.

Issue #65, opened 2026-09-14
K1.2: the storage family skeleton, with repetitions, warm-up, statistics and a document schema


spdrman:
The PMTiles harness in the engine repository measures one process per cell, once. A cell carries a
single value copied into a field called `median`, no warm-up, no dispersion, and no way to tell a
real move from the noise floor. Both sweep profiles walk `2048x2048@256`, so the committed exports
contain a free replicate pair on an idle host with identical code, and across it p99 moves 74.5%,
wall 10.7% and p50 7.1%. Nothing in the suite can see that.

Build the `storage` family in libviprs-bench on the repetition and isolation machinery this crate
already owns (child per cell with `wait4` `ru_maxrss`, seven-plus timed iterations after a warm-up,
median, IQR and CI95, interleaved order) and give it a document schema that cannot express a
single-shot number as a median.

## Shape

`src/storage/{mod,cells,scenarios,document,stats}.rs`, a `storage` bin, a `storage` Dockerfile stage
from the pinned base that skips the libvips build entirely, and `run-bench.sh storage`. Every read
scenario runs in its own process on a fresh reader, because the current harness lets `read_random`
walk a reader `read_sequential` just warmed and then reports it as a cold-ish random read.

Unmeasured columns are `null`, never `0`. A real `filesystem_entries: 0` is exactly the failure the
null rule exists to expose, so the migration from an old export is per scenario and explicit.

## Tests first

- `every_read_scenario_runs_in_its_own_process_on_a_fresh_reader`: wrap the archive in the counting
  `RangeReader` the test suite already has, run `read_random` after `read_plan_order`, assert the
  first leaf-addressed lookup issued a directory range read. RED against reusing the warmed reader.
- `a_cell_carries_reps_samples_and_a_median_of_them`: `samples.len() == reps`, `median ==
  median(samples)`, `ci95[0] <= median <= ci95[1]`. RED against a single-shot cell copying its one
  value into `median`.
- `generate_reps_regenerate_and_the_artefact_digest_does_not_move`: each rep's `filesystem_entries`
  and `artefact_digest` equal, its tempdir different. RED against generating once and timing nothing
  seven times.
- `pass_scenarios_discard_one_warmup_and_record_it`: assert `warmup.policy` is recorded and the
  first pass's value is absent from `samples`. RED against measuring the first pass.
- `unmeasured_columns_are_null_never_zero`. RED against a zero-filled row.
[...]

PR #68, opened 2026-09-14
K1.3: provenance, the emulation probe, attestation, digests and the archive

### LINKED ISSUE #66: K1.3: provenance, the emulation probe, attestation, digests and the archive
`docs/pmtiles-benchmarks.md` publishes 480 lines of numbers and records no platform at all. The
issue that produced them describes its own run as being in "the amd64 Linux container", this Mac is
Apple Silicon, `DOCKER_DEFAULT_PLATFORM` here is `linux/amd64` and Docker Desktop has Rosetta on, so
those numbers were almost certainly emulated. Nothing in the artefact can settle it, and that
silence is the defect rather than the emulation.

`libviprs_bench::provenance::Provenance` already carries commit, host, ncpu, load average, rustc,
thermal and container flag, with `measurement_condition_warnings()` flagging load >= ncpu. It was
simply never applied to the storage question. Extend it, add the one piece nobody has written (an
emulation probe), and put an archive behind it that refuses a run rather than averaging it in.

## The probe is the whole point

A probe that always answers `false` passes every Rust unit test you could write for it. The positive
control is on this machine: build it, run it under `--platform linux/amd64` on this arm64 host and
expect `true`, run it under `--platform linux/arm64` and expect `false`. That is a shell test, not a
cargo test, and it is the thing the finding rests on.

Whether `/proc/sys/fs/binfmt_misc` is visible inside a Docker Desktop container, and what
`/proc/self/maps` shows under Rosetta, are unknown and are what the positive control exists to
measure. If both are blind, fall back to comparing the runner-supplied daemon architecture, and make
the document say which evidence it used.

## Extend, attest, archive

Provenance gains: emulation, filesystem type of the scratch directory, cgroup limits, RUSTFLAGS,
`debug_assertions`, both trees' commit and dirty flag, and `Cargo.lock` dependencies via `build.rs`
and `cargo metadata`. Then `src/storage/{attest,integrity,archive}.rs` and a `storage-aggregate`
bin that refuses rather than reports.

A null commit and a true dirty flag are easy to produce by accident, and the #1022 campaign hit
three ways: a linked worktree bind-mounted into a container has a `.git` file pointing outside the
mount so git answers "not a git repository"; a tree on a NAS share read as root trips
`safe.directory` so every git call exits 128; and that share forcing mode 0777 makes all 814 files
read as mode changes so `dirty` comes back true on a clean tree. All three are fixed from the
environment, and all three argue for the stderr warnings rather than against them.

## Tests first

- `the_emulation_probe_answers_true_under_amd64_on_this_arm64_host`: the shell positive control
  above. RED against a probe that always says `false`.
- `an_emulated_or_unknown_document_is_refused_by_the_aggregator`. RED against a rule that only
  refuses `true`.
- `a_document_without_commit_dirty_or_invocation_is_refused`;
  `a_dirty_tree_is_refused_without_the_flag_and_stamped_with_it`;
  `resolved_reps_must_equal_every_cells_reps`; `a_debug_build_is_refused`: each RED against an
  aggregator that accepts.
- `storage_attestation_is_observed_not_asserted`: feed a cell declaring regime `root` and an archive
  whose `root_entries` are leaf pointers, assert refused. RED against an attestation that reads the
  cell's own label.
- `run_id_is_derived_from_the_document_not_the_clock`: archive the same document twice, assert one
  id. RED against `SystemTime::now()` in the id.
- `verify_recomputes_all_four_digests_and_names_the_block_that_moved`: edit one sample, assert
  `cells` and `document` moved and `runners` did not.
- `the_scratch_filesystem_is_recorded_and_tmpfs_is_refused_by_default`.

The cross-language digest test against the JavaScript reference belongs to K2.2, but pin the
canonicalisation rules (key order, float formatting, absent versus null) here and write them down,
because reconciling them late costs an afternoon.

## Not in scope

Scenarios and cells (K1.4), the fitted gate (K2.3), anything on the site.

spdrman:
Closes #66

`docs/pmtiles-benchmarks.md` publishes 480 lines of numbers and records no platform at all. The run
that produced them calls itself "the amd64 Linux container", this Mac is Apple Silicon,
`DOCKER_DEFAULT_PLATFORM` here is `linux/amd64` and Docker Desktop has Rosetta on, so those numbers
were almost certainly instruction-translated. Nothing in the artefact can settle it, and that silence
is the defect rather than the emulation.

So: a benchmark document that can say what produced it, and an aggregator that refuses a run it
cannot vouch for instead of averaging it in.

## The probe, and the control that can fail

A probe that always answers `false` passes every Rust unit test anyone could write for it, so the
evidence is `tools/probe-emulation.sh`. It compiles `src/probe_emulation_main.rs` with bare `rustc`
under `--platform linux/amd64` and again under `--platform linux/arm64` on one machine and asserts
opposite answers. Bare `rustc` rather than cargo because the probe is `std` only, and building it
through cargo would drag the whole `libviprs` path dependency through an emulated compiler for two
hundred lines of `/proc` reads.

Two things were genuinely unknown before I ran it, and the control is what measured them:

| source | `--platform linux/amd64` | `--platform linux/arm64` |
|---|---|---|
| `/proc/sys/fs/binfmt_misc` | present and **empty** | present and empty |
| `/proc/self/maps` | carries `/run/rosetta/rosetta` | clean |
| `BENCH_DAEMON_ARCH` against binary arch | `arm64` against `x86_64` | `arm64` against `aarch64` |

`binfmt_misc` is blind inside a Docker Desktop container: the directory exists and lists nothing
whether or not Rosetta is translating the very process reading it, so an empty listing is exactly as
informative as an unreadable one and the probe never returns a native verdict from it.
`/proc/self/maps` is not blind, and it is the primary evidence. Rosetta maps its translator into
every process it translates and the path survives even though `/run/rosetta` cannot be opened through
the container's mount namespace (`ls -d /run/rosetta` says no such file in the same container where
the mapping is right there in maps). The runner-supplied daemon architecture is the fallback for a
host where `/proc` says nothing, and a third run with `BENCH_DAEMON_ARCH` withheld proves the maps
evidence carries the verdict on its own here.

The document records which evidence it used, per source, not only what it concluded.

## The canonicalisation rules, pinned now

K2.2 will pin a cross-language test against causl's JavaScript `digest.mjs`. I do not write that test,
but the rules it will hold both sides to are decided here, because a float format reconciled after
both sides have shipped archives invalidates every digest already written down. The reference is
`causl-bench/tools/suite/regression-gate.mjs`, `canonicalJson` at 1502 and `baselineDigest` at 1529.

1. **Key order.** Keys ascending, arrays untouched. JavaScript sorts by UTF-16 code unit and Rust by
   UTF-8 byte, and those orders differ: `"\u{10000}"` is the surrogate pair `D800 DC00`, so
   JavaScript puts it before `"\u{FFFD}"` and Rust puts it after. They agree on ASCII, so a non-ASCII
   key is **refused** rather than digested two ways.
2. **Absent is not null.** An absent key is not emitted, an explicit `null` is emitted as `null`, and
   the two are different documents with different digests. The consequence for a producer is the
   load-bearing one: **no `#[serde(skip_serializing_if)]` anywhere in a storage document**.
3. **Numbers.** A value with no fractional part prints as a plain decimal integer whatever Rust type
   it arrived in, because `JSON.stringify(1.0)` is `"1"` and `serde_json` writes `1.0`. `-0.0` prints
   as `0`. Everything else prints as the shortest decimal that round-trips, which is the same digit
   string both languages produce *in plain notation*. Outside that range they diverge, JavaScript
   writing `1e+21` and `1e-7` where Rust writes the digits in full, so any value at or above `1e21`
   or strictly below `1e-6` is refused, as is any integer beyond `2^53-1` and anything non-finite.
   Those two thresholds are ECMA-262's, not mine.
4. **Strings.** RFC 8259 escaping: `\"`, `\\`, the five two-character control escapes, `\u00xx` for
   the remaining C0 controls, everything else literal, `/` and non-ASCII unescaped.

A digest is `sha256:` plus 64 lowercase hex over the UTF-8 bytes of the canonical string. The
`document` digest covers the whole document with `integrity` and `combinedAt` removed, so sealing does
not change what was sealed and `--verify` is a check that can pass.

## Attestation is observed, never asserted

A cell earns `storageAttested: true` from two observations, neither of which reads the cell's label:
the archive's root directory is walked and the regime it is *in* is compared against the regime the
cell *declares*, and a seeded sample of at least 64 coordinates is read back from both backends and
compared byte for byte. An empty root directory is an unobserved archive rather than a small one,
which is the difference between an answer and a code path that looked at nothing.

## The three ways a null commit happens by accident

All three hit for real in the previous campaign, none of them is the harness's fault, and all three
are fixed from the environment. `build.rs` detects each and names it in a `cargo:warning=` rather than
swallowing it, and the note travels into the document so the refusal can say which case it is in:

1. A linked git worktree bind-mounted into a container has a `.git` that is a *file* pointing at a
   gitdir outside the mount, so git answers "not a git repository". This branch is being developed in
   one, so I hit it immediately. Detected by reading the pointer and checking whether its target
   exists, which distinguishes it from a tarball with no repository at all; fixed by mounting the real
   gitdir at the same absolute path, which is what the lane's container command does.
2. A tree that trips `safe.directory` exits 128 on every git call. Cleared with
   `-c safe.directory=<abs path>` for the one directory being asked about, not the blanket `*`.
3. A share forcing mode 0777 makes every file read as a mode change, so a tree nobody edited comes
   back dirty. `git status` is run both with and without file-mode comparison and the difference is
   reported, rather than quietly masked.

Case 1 is not hypothetical here, and I ran it rather than reasoned about it. The same test suite in a
container with the gitdir deliberately not mounted produces:

```
warning: libviprs-bench@0.3.0: provenance: harness tree at /Users/rom/workspace/libviprs/.epick/_wt/k13
 -- this is a linked git worktree whose .git file points at
 /Users/rom/workspace/libviprs/libviprs-bench/.git/worktrees/k13, which does not exist here;
 bind-mount that directory at the same absolute path and the commit comes back;
 git does not see a repository here (git rev-parse HEAD exited 128: fatal: not a git repository: ...)
```

which is the difference between a null commit somebody can fix in thirty seconds and a null commit
they spend an afternoon on.

## The digest came from a parse, and it should not have

K1.2 found this from the other side. I checked it before building on it, and it is real and slightly
worse than reported.

`serde_json`'s number **reader** is not correctly rounded by default. In an arm64 container, on the
witness `0.09090909090909091`, which is the coefficient of variation of `[10, 11, 12]` and therefore a
value this suite actually produces:

```
witness text          0.09090909090909091
std parse             3fb745d1745d1746   prints 0.09090909090909091
serde_json parse      3fb745d1745d1747   prints 0.09090909090909093
serde_json print(std) 0.09090909090909091
computed 1.0/11.0     3fb745d1745d1746   prints 0.09090909090909091
printed then reread   3fb745d1745d1747   ROUND TRIP EXACT = false
```

The printer is right and `std` agrees with it, so the reader is the one that is wrong. The last two
lines are the part that had not been stated: **the round trip is not exact within Rust either**, so a
document whose bytes are exactly what its producer wrote failed its own integrity check. `--verify`
read the file with `from_str`, re-canonicalised the floats that came back, and refused a file with
nothing wrong with it. Across languages it is worse and I cannot see it from here: V8's `JSON.parse`
**is** correctly rounded, so Rust and JavaScript read one archived file as two different floats and
K2.2's cross-language test would go red with neither implementation at fault.

### What I did first, and why I changed it

My first fix was a JSON reader in `integrity.rs` that kept every number as the text it was written as,
so nothing on the digest path built a float. It worked, it had 33 red mutations behind it, and it
reduced the digest to bytes, which is the strongest guarantee available.

The coordinator pointed out that serde_json ships **`float_roundtrip`**, which I had not known about,
and asked for the comparison on the record rather than a rebuild on their say-so. So I ran it: feature
on, bespoke reader off the digest path, every test kept.

```
PROBE std        = 3fb745d1745d1746
PROBE serde_json = 3fb745d1745d1746
PROBE equal      = true
PROBE roundtrip  = true

test a_document_written_and_read_back_verifies_against_its_own_bytes ... ok
test result: FAILED. 24 passed; 1 failed
```

The one failure was `printing_a_float_and_reading_it_back_through_serde_json_is_not_the_identity`,
failing **because its premise is that serde_json is broken**. Its own message said that if it ever
passed the feature had been fixed and it should be retired rather than relaxed, and that is what
happened.

Two other tests passed but only because they called the reader's own API directly, so they would have
gone with it: the duplicate-key refusal and the escape-decoding test.

### The comparison, and what I chose

**I took `float_roundtrip` and deleted the reader.** `integrity.rs` goes from 1063 lines to 537.

| | bespoke reader | `float_roundtrip` |
|---|---|---|
| fixes the round-trip bug | yes | **yes**, proven by the run above |
| new crates in the graph | none | **none**, it gates an in-tree `lexical` module (`float_roundtrip = []` in serde_json 1.0.151) |
| API or feature interactions | none | **none**; unlike `arbitrary_precision` it does not touch `Value`, `to_value`, or `preserve_order` |
| lines of hand-written parsing under the trust module | ~450 | **0** |
| guaranteed by | reading the code | a feature flag, **with a test as its canary** |

The last row was the one I had to think hardest about, and it resolved in favour of the feature:
`the_json_reader_is_correctly_rounded` fails in one line if the flag is removed, and
`a_document_written_and_read_back_verifies_against_its_own_bytes` fails with it. That is a stronger
guard than "nobody edits this file", and both are in the mutation table below with the flag removal as
the mutation.

**The specific thing I was asked to weigh: I am not cashing in the reader's headline advantage.** The
`>= 1e21` and `< 1e-6` refusals stay either way, because they exist for a *printer* disagreement
(JavaScript writes `1e+21` and `1e-7` in exponent form where Rust writes the digits out) and that is
present whether the canonical form is derived from a token or from a float. The reader never let me
drop them, so "it never has to reason about how a float prints" was not true of it. That observation
is most of why the comparison went the way it did.

**What the reader bought that the feature does not, stated honestly, because none of it was
load-bearing for a test:**

- *Duplicate-key refusal.* `serde_json` keeps the last (`{"a":1,"a":2}` reads as `{"a":2}`), so the
  refusal is unreachable without a reader of my own. I checked whether it was a cross-language hazard
  before mourning it: **V8's `JSON.parse` also keeps the last**, so both languages resolve the
  ambiguity the same way and the digests agree. It was a nice-to-have the reader made free, not
  something the contract needs. The test went with it and the capability loss is recorded here.
- *Byte preservation on re-seal.* With the feature on, parse then print is exact, so a machine-written
  file re-seals to the same tokens; only a hand-written non-shortest token like `1.10` would be
  rewritten (to `1.1`, the same value, and the canonicaliser normalises trailing zeros anyway). No
  test proved it mattered.
- *A structural rather than a flag-shaped guarantee.* Real, and answered by the canary above.

One thing the reader would genuinely have given, for completeness: for a document **written by
JavaScript**, emitting its tokens verbatim makes Rust and JS agree unconditionally. Without it, both
sides re-derive digits and agreement rests on Rust's `Display` and V8's `Number::toString` producing
the same shortest string. But for a **Rust-written** document, which is every storage document, the
reader's version of that dependency was ryu against V8 rather than `Display` against V8: two
assumptions of identical shape. It is K2.2's to pin either way, and I have flagged it there.

### Which of my existing digest tests would have caught the original bug

**Neither direction. Not one.** Every digest test I had canonicalised a `Value` built in memory with
`json!` and compared it against another `Value` built in memory. None went near a file, so none
exercised a parse, and all 23 stayed green while the new round-trip test failed. The tests were about
the canonicaliser and the bug was in the reader underneath it.

### Red then green

```
# the branch as it stood, --verify's own from_str + Value path, feature off
test a_document_written_and_read_back_verifies_against_its_own_bytes ... FAILED
test result: FAILED. 22 passed; 1 failed          exit 101
# with float_roundtrip on
test a_document_written_and_read_back_verifies_against_its_own_bytes ... ok
test result: ok. 24 passed; 0 failed              exit 0
```

End to end through the binary, which is where it bit (run against the byte-path build; the
observable behaviour is identical on the shipped one):

```
storage-aggregate --check   run.json  -> admissible                                exit 0
storage-aggregate --archive run.json  -> archived as 20260913T214500Z-0f1e2d3c...   exit 0
grep -o '"cov": [0-9.]*' archive/<run>.json  ->  "cov": 0.09090909090909091
storage-aggregate --verify  archive/<run>.json
  verified: all four digests still hold (document sha256:d67ce9a8...)              exit 0
storage-aggregate --archive run.json  -> already archived, digest still matches     exit 0
sed -i 's/10\.0,/10.5,/' <run>.json; storage-aggregate --verify <run>.json
  REFUSED: the cells block moved. / REFUSED: the document block moved.
  the runners and measurements block(s) did not move                               exit 1
```

### The general form

`parse(print(x)) == x` was the assumption the bug was made of. It is now true, by a feature flag with
a canary, and the places that relied on it are named: `verify`, `seal` and `archive` all read files,
and the `Value` entry points are documented as correct for a value the caller **constructed**.

## SHA-256: taken (`46c796f`)

I wrote it by hand because `[dependencies]` was not this lane's to edit. That reason went away, and
checking rather than assuming turned up the thing that settles it: **`sha2` 0.11 is already a normal
dependency of `libviprs`**, which is a normal path dependency here, so every binary in this crate was
already compiling it. Promoting it from dev-dependency to dependency adds **zero** crates to anything.
That turns a real trade about what enters the normal graph into 197 lines of hand-written hashing in
the one module whose job is to be trusted, bought for nothing.

`src/sha256.rs` is now a 30-line wrapper plus the vectors. `build.rs` cannot use its own crate's
library, so `sha2` is listed again under `[build-dependencies]` for the `Cargo.lock` digest: one extra
compilation for the host that never reaches a shipped binary, which is the entire cost.

**The vectors are carried over unchanged and now point at what ships.** They earned that. Written
against the hand-rolled version, they are what found that the three published FIPS 180-4 examples plus
the 55/56/63/64/65-byte padding walk top out at **two blocks** and are **all ASCII**, so between them
they never exercised the chaining across a third block, never put a byte above `0x7f` through the word
packing, and never pushed the length field past 16 bits. Documents digested here run to hundreds of
blocks, so that was a real hole. `long_and_high_bit_messages_come_out_right` adds 1000 bytes (16
blocks), 100000 bytes (1563 blocks) and all 256 byte values four times over, checked against
`sha256sum` and Python's `hashlib`, which agreed with each other.

**Still uncovered, and written into the test:** a message at or beyond 2^32 bits (512 MiB), where a
length field truncated to 32 bits would go wrong and which no document this suite produces reaches.

What the vectors guard changed with the implementation, and it is worth being plain about: against the
hand-rolled version they guarded the compression function, and three mutation rows reddened inside it
(`sha256-constant`, `sha256-padding`, `sha256-only-one-block`). Those rows are retired with the code
they mutated. Against `sha2` the vectors guard the **wiring** — the hex formatting, the byte order,
that the right function is called at all — because the compression function is upstream's and has its
own suite. `sha256-nibble-swap` is the row that now holds that, and it reddens all three vectors.

## The parked core: a warning, and the obvious predicate would have missed it (`f899e9c`)

Not a refusal. Benchmarking a core that is not on `main` is a legitimate thing to want, an intentional
baseline looks exactly like this, and provenance's job is to make a run's conditions legible rather
than to narrow what may be measured. What was wrong was only that it passed silently.

**I measured the named test case instead of implementing against it, and the literal predicate stays
silent on it.** The check as specified is "warn when the core commit is not an ancestor of the core's
`origin/main`". On this machine:

```
$ git -C libviprs merge-base --is-ancestor HEAD origin/main; echo $?
0
$ git -C libviprs rev-list --count HEAD..origin/main   # 458
$ git -C libviprs rev-list --count origin/main..HEAD   # 0
```

`576910df` **is** an ancestor of `809ee801`. The core is 458 commits *behind* main, not diverged from
it, so an ancestry test would have said nothing about the very tree it was written for.

So there are two conditions and they get different sentences. **Diverged** means the commit is not in
`origin/main`'s history at all, so the run measured work that is not in the mainline. **Behind** means
it is in that history and main has moved on, which is what an intentional baseline looks like and
equally what a checkout somebody forgot to update looks like. Both warn; `at-main` is silent, because
a warning that fires on a clean tree is one nobody reads; and when there is no local `origin/main` it
says it could not tell rather than implying the tree is fine.

**Entirely local.** `origin/main` is `refs/remotes/origin/main`, a ref already in the tree, so the
build script does not fetch. The cost is that the answer is only as fresh as the last fetch, which is
written into the code comment; for a warning rather than a gate that is the right trade.

Fires at build time and again at run time, and the document carries the condition so it survives into
the archive rather than living in a terminal somebody has closed:

```
warning: libviprs-bench@0.3.0: provenance: the measured libviprs is behind its own origin/main.
  HEAD is 576910dff04567274ed88b4c199a45bb04acc8d2, origin/main is 809ee8014d002518ce55edaceba698ca7a8b8a79,
  which is 458 commits ahead of HEAD. This is not refused, ...

$ storage-aggregate --provenance
  "library": { "commit": "576910df...", "mainCommit": "809ee801...",
               "mainRelation": "behind", "commitsBehindMain": 458, ... }
WARNING: the measured libviprs is behind its own origin/main. ...
```

The harness gets its relation recorded but no warning: it is on a lane branch every day of the week,
and the library is the thing being measured.

## The core working tree, after K1.4's gradient## The core working tree, after K1.4's gradient

Checked, and the answer is that my one read of it is deliberate. `build.rs` reads `../libviprs`'s
manifest and git state, and it must: cargo links that tree as a path dependency, so the provenance has
to describe *it*. Reading `origin/main` instead would be a lie about what was measured.

The parked branch made that visible rather than hiding it. The core checkout is on `m826` at
`576910df`, and `origin/main` is `809ee801`; `--provenance` records:

```
library: {"commit": "576910dff04567274ed88b4c199a45bb04acc8d2", "dirty": false,
          "gitNote": "clean read", "name": "libviprs", "version": "0.4.0"}
```

which is the parked branch, correctly. Nothing of mine reads a source file or a fixture from that tree
through the working copy, so K1.4's trap does not reach this lane: their failure was needing canonical
content and getting a parked branch, and mine is the inverse.

Worth one sentence for whoever composes: an archived run taken while the core is parked off `main`
will faithfully record a commit that is not on `main`, and nothing refuses that. The commit is in the
document and in the run id's environment bucket, so it is visible rather than silent, and §5.4 does
not ask for more. Flagging it rather than adding a rule.

## Mutation table

Every row below was run against the tree as it now stands, one mutation at a time, and **every row
required a `Compiling libviprs-bench` line in its own build output before a verdict was recorded**.
A row that compiled nothing proves nothing and is the same colour as one that genuinely passed, so
the runner also stamps each mutated file's mtime forward rather than hoping to miss the timestamp
granularity, retries once if the witness is absent, and records `INCONCLUSIVE` rather than a colour if
it is still absent. No row came back INCONCLUSIVE and none needed the retry.

**32 rows, all red.** The published table was re-run rather than reasoned about, per the
coordinator's request after K1.2 caught a green row that was a stale build. Six rows were re-targeted
because the `float_roundtrip` adoption replaced the code the old anchors pointed at, and five were
retired with the reader they tested. Three more were retired with the hand-rolled SHA-256's internals
and three added, all of which were run the same way.

| test | mutation | observed |
|---|---|---|
| `the_emulation_probe_answers_true_under_amd64_on_this_arm64_host` (`tools/probe-emulation.sh control`) | the probe answers `false` unconditionally | **red**: `FAIL: linux/amd64 reported emulated=false, expected true`, exit 1. The `linux/arm64` side still passed, which is the point: one mutation, one side. This row cannot go stale: the control builds with bare `rustc` in a fresh container with no cache at all |
| `the_published_vectors_come_out_right` | `sha256_hex` emits each byte's nibbles in the wrong order, so the wiring around `sha2` is wrong while `sha2` itself is right | **red** (reddens all three vector tests; it guards the wiring around `sha2`, not the compression function, which is upstream's) |
| `a_core_that_is_not_at_main_warns_and_is_still_archivable` | a core that is not at its own origin/main passes silently, which is the condition before this commit | **red** |
| `a_core_that_is_not_at_main_warns_and_is_still_archivable` | only a DIVERGED core warns, which is the check as literally specified and is silent on the tree it was specified for | **red**, which is the row that says the predicate as specified is silent on the real tree |
| `a_substring_match_on_the_whole_path_is_not_the_rule` | `path_is_translator` becomes a substring match on the whole path | **red** |
| `docker_and_rust_arch_spellings_are_reconciled_before_they_are_compared` | `normalise_arch` returns its input unchanged, so `amd64` never equals `x86_64` | **red** |
| `an_empty_binfmt_misc_directory_is_not_evidence_of_anything` | an empty `binfmt_misc` directory reads as a native verdict | **red** |
| `unknown_is_not_spelled_false` | `Emulated::Unknown` serialises as `false` | **red** |
| `verify_recomputes_all_four_digests_and_names_the_block_that_moved` | `verify` stops at the first block that moved | **red** |
| `an_emulated_or_unknown_document_is_refused_by_the_aggregator` | only `emulated: true` is refused, so `"unknown"`, `null` and absent all pass | **red** |
| `a_document_without_commit_dirty_or_invocation_is_refused` | the missing-commit refusal is switched off | **red** |
| `a_dirty_tree_is_refused_without_the_flag_and_stamped_with_it` | `--allow-dirty` no longer requires the stamp on every cell | **red** |
| `resolved_reps_must_equal_every_cells_reps` | only the first cell's reps are compared against the resolved value | **red** |
| `a_debug_build_is_refused` | a build with debug assertions on is accepted | **red** |
| `the_scratch_filesystem_is_recorded_and_tmpfs_is_refused_by_default` | tmpfs is accepted without the profile declaring it | **red** |
| `an_ok_cell_without_attestation_is_refused` | an ok cell no longer has to be attested | **red** |
| `the_clean_fixture_is_admitted` | the tmpfs rule is inverted, so every non-tmpfs filesystem is refused | **red** (`the_scratch_filesystem_is_recorded_and_tmpfs_is_refused_by_default` reddens too) |
| `run_id_is_derived_from_the_document_not_the_clock` | the wall clock is appended to the run id | **red** |
| `storage_attestation_is_observed_not_asserted` | the attestation reads the cell's declared regime instead of walking the archive | **red** |
| `the_dependency_graph_names_the_measured_library` | the resolved dependency graph comes back empty | **red** |
| `the_toolchain_block_is_read_from_the_build_not_declared` | cargo's version becomes a hard-coded placeholder instead of a build stamp | **red** |
| `the_probe_control_asserts_both_directions` | the control stops asserting the foreign platform's answer | **red** |
| `an_integral_float_digests_as_javascript_prints_it` | `canonical_json` delegates to `serde_json::to_string`, which writes an integral float as `1205.0` where JavaScript writes `1205` | **red** |
| `absent_and_null_are_different_documents` | null-valued keys are dropped, which is what `skip_serializing_if` does to a document | **red** |
| `a_number_javascript_would_print_in_exponent_form_is_refused` | the plain-notation range check is removed and the number is just printed | **red** |
| `a_non_ascii_key_is_refused_because_the_two_sort_orders_disagree` | keys are sorted without the ASCII restriction | **red** |
| `sealing_a_document_does_not_change_what_was_sealed` | the document digest stops excluding `combinedAt`, so sealing changes what was sealed | **red** |
| `the_archive_refuses_rather_than_reports` | the archive files the document before checking whether it is admissible | **red** |
| `escapes_and_layout_do_not_change_a_digest` | `canonical_string` stops escaping C0 control characters, so the canonical form is not what JSON.stringify writes | **red** |
| `the_json_reader_is_correctly_rounded` | `float_roundtrip` is taken off serde_json, which is the whole fix | **red** |
| `a_document_written_and_read_back_verifies_against_its_own_bytes` | `float_roundtrip` is taken off serde_json, so a document digests differently from the bytes its producer wrote | **red** |
| `the_build_stamps_name_both_trees_and_say_why_when_they_cannot` | the tree note is hard-coded to "clean read", so a tree git could not read claims it was read cleanly | **red**, run in a container with the gitdir NOT mounted, which is the linked-worktree trap |

## What I ran, and what came out

All of it in `linux/arm64` containers, `--platform` spelled out on every `docker build` and
`docker run`, never the host toolchain. The probe control is the deliberate exception: it runs both
platforms, because that is the whole point of it.

```
$ ./tools/probe-emulation.sh control
probe-emulation: daemon arch arm64 (aarch64)
probe-emulation: expecting emulated=false under linux/arm64 and true under linux/amd64
PASS: linux/arm64 reported emulated=false
PASS: linux/amd64 reported emulated=true
PASS: linux/amd64 reported emulated=true from /proc evidence alone
probe-emulation: both required assertions hold on this machine
EXIT=0
```

And with the probe mutated to answer `false` unconditionally, which is the implementation no cargo
test can catch:

```
PASS: linux/arm64 reported emulated=false
FAIL: linux/amd64 reported emulated=false, expected true
probe-emulation: 1 of the two required assertions failed
EXIT=1
```

The whole suite, in one arm64 container: `cargo test --no-fail-fast` is **139 passed, 0 failed,
3 ignored** across 27 test binaries, `cargo clippy --lib --bins --tests -- -D warnings` exits 0 and
`cargo fmt --check` exits 0.

One thing worth writing down because it cost me a wrong conclusion for ten minutes: three
`tests/version_matrix.rs` cases do `git worktree add` in the sibling core repo, and my lane's
container first mounted that repo read-only, so they failed. That is the mount and not this branch,
and I checked it rather than assuming it: with the core mounted read-write,
`checkout_resolves_head_sha_into_a_temp_worktree_and_cleans_up` passes and so does everything else.

## Folded in rather than filed

`provenance::cpu_model` returned `"unknown"` on aarch64 Linux, because `/proc/cpuinfo` has no `model
name` line there. That is not cosmetic: the run id hashes the cpu model into its environment bucket,
so every arm64 host in the world would have landed in one bucket and two incomparable runs would have
looked comparable. It now falls back to `/proc/device-tree/model` and then to the implementer, part
and architecture fields.

`SUITE-PLAN.md` §5.4 asks for the dependency graph "via `build.rs` and `cargo metadata`". `build.rs`
parses `Cargo.lock` directly instead, and the reason is narrow: a build script that shells out to
`cargo metadata` runs cargo underneath a cargo that may be holding the package-cache lock, which is a
documented way to wedge a build until somebody kills it. `cargo metadata` resolves *from* `Cargo.lock`,
so the lockfile is the same facts by a route that cannot deadlock, and `source` and `checksum` are
Cargo's own spellings of npm's `resolved` and `integrity`.

SHA-256 is written out by hand in `src/sha256.rs` rather than added as a dependency. `[dependencies]`
is not this lane's to edit, a build script cannot use its own crate's library so `build.rs` would have
needed a second copy under `[build-dependencies]` anyway, and this is a content digest for archived
documents rather than a security boundary. Its correctness rests on the FIPS 180-4 vectors and on five
block-boundary lengths, all cross-checked against `sha256sum`, not on it compiling.

## Coordination

The `provenance` block's exact shape is posted as a comment on #65 for K1.2, since it fills a slot in
their document type. I touched `Cargo.toml` only to add two `[[bin]]` entries, and `src/storage/mod.rs`
declares only this lane's three modules with a comment saying whose lines go next to them.

## Not proven

I did not build the whole crate under `--platform linux/amd64` to watch the aggregator refuse a live
emulated run end to end. Four lanes are building at once and that is a slow emulated compile of
`libviprs` for a demonstration the probe control and the unit tests already make. The refusal of
`emulated: true`, `"unknown"`, `null` and absent is proven by test, and the probe's two-sided answer
is proven by the control.


spdrman:
Four things I found while building this and folded into the branch rather than filing, since the
campaign rule is to fold and say so.

**`cpu_model` returned `"unknown"` on aarch64 Linux.** `/proc/cpuinfo` has no `model name` line on
arm, so the existing scan fell through. That is not cosmetic here: the run id hashes the cpu model
into its environment bucket, so every arm64 host would have landed in one bucket and two runs on
completely different machines would have looked comparable. It now tries
`/proc/device-tree/model` first and then composes from the implementer, architecture, part and
variant fields. On this Mac's containers it reads
`aarch64 CPU implementer=0x61 CPU architecture=8 CPU part=0x000 CPU variant=0x0`.

**The history golden had to move.** `tests/chart_shape_drift.rs` pins the serde shape of a
`BenchmarkSnapshot` against `tools/charts/fixtures/golden_history.json`, so extending `Provenance`
drifted it. That test did exactly what it exists for, and its own failure message names the two files
to update. I updated the golden; `tools/charts/render.mjs` reads no provenance key, so it needed
nothing. Flagging it because it is a file outside this lane's stated scope.

**`cargo metadata` in `build.rs` is not what landed.** §5.4 asks for the dependency graph "via
`build.rs` and `cargo metadata`". A build script that shells out to `cargo metadata` runs cargo
underneath a cargo that may be holding the package-cache lock, which is a known way to wedge a build
until somebody kills it. `cargo metadata` resolves *from* `Cargo.lock`, so `build.rs` parses the
lockfile instead: same facts, no deadlock, and `source` and `checksum` are Cargo's own spellings of
npm's `resolved` and `integrity`. Worth a second opinion if anyone disagrees.

**SHA-256 is written out by hand** in `src/sha256.rs` rather than promoted from a dev-dependency.
`[dependencies]` is not this lane's to edit, a build script cannot use its own crate so `build.rs`
would have needed a second copy under `[build-dependencies]` anyway, and this is a content digest and
not a security boundary. Its correctness rests on the FIPS 180-4 vectors plus five block-boundary
lengths, all cross-checked against `sha256sum` in a container, and both are in the mutation table.

One coordination note: `src/storage/mod.rs` exists on this branch declaring only this lane's three
modules, and K1.2 and K1.4 each have their own version of that file. It is a three-line conflict that
resolves by taking all of them. The overlap worth a real look at compose time is K1.2's
`archive_shape` against `storage::attest::ObservedArchive`; I have raised that on #65.

spdrman:
The compose landed here: `4b8ef84` merges `lane/k1.4-scenarios`, which carries K1.1 and K1.2 with it. The
branch now contains all four lanes (checked with `merge-base --is-ancestor` rather than taken on
trust), and the merge base was `origin/main`, so nothing came in twice.

## Conflicts: three, not the four expected

`Cargo.toml`, `src/lib.rs` and `src/storage/mod.rs`.
`tools/charts/fixtures/golden_history.json` auto-merged, and I verified that rather than trusting it,
because a JSON auto-merge is the kind of thing that looks fine and is not: `chart_shape_drift` passes,
which is the test that pins the serde shape it exists for.

**`serde_json` takes both features**, `["preserve_order", "float_roundtrip"]`, and both are
load-bearing. Two mutation rows prove it independently: removing only `preserve_order` reddens K1.2's
`the_document_round_trips_with_field_order`, and removing only `float_roundtrip` reddens both of my
canaries. **`sha2`** is listed once in `[dependencies]`, with the `[build-dependencies]` entry kept for
`build.rs`. `src/storage/mod.rs` takes K1.4's side as the base with this lane's three modules declared
into it.

## The SHA-256 collapse, measured rather than argued

Converged on `src/sha256.rs`, which grows a streaming form so `artefact_digest` can keep hashing an
artefact a buffer at a time. Both forms now share one hex formatter: the merge found two wrappers
around the same crate, each with its own hex loop, and nothing checking they agreed. K1.4's
tree-digest definition is untouched.

That collapse could have moved `artefact_digest` **on every architecture at once**, which is exactly
what a cross-architecture comparison cannot see. So I measured it. Five digests captured on the merged
tree before the collapse and again after, including a 3 MB file that crosses the 1 MiB read buffer and
the tree digest itself:

```
a.bin          e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
b.bin          ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad
z/c.bin        6d1cf22d7cc09b085dfc25ee1a1f3ae0265804c607bc2074ad253bcc82fd81ee
z/inner/d.bin  1913233a0a87fe912497ee543021c40adc5d414614fc76fdff3e0c08b6a1d981
tree           49ae1d007e6f05625180663add937b87d73b3f72505187d102c0ef3f1b0a84bb
$ diff before after   ->  identical
```

They are now pinned in `artefact_digest_is_pinned_to_the_values_the_two_architectures_agreed_on`, so
the next change to the hashing path has to answer for them instead of rediscovering the question. The
mutation row `artefact-digest-moved` feeds the hasher one byte short of each buffer, which is precisely
the change that moves the digest identically on both architectures.

## K1.2's parser test, closed the way its author said to

`serde_jsons_number_parser_is_not_correctly_rounded` goes red under `float_roundtrip`, because its
premise is that the defect exists. Its own note: *"If serde_json ever closes this, the assert_ne! below
goes red. That is the signal to tighten the round trip above to byte identity and delete this."* Done,
both: `the_document_round_trips_with_field_order`'s `cov` comparison is now `to_bits()` equality rather
than a `1e-15` tolerance, and the test is gone, with the comment pointing at the canary that replaces
it.

## Mutations after the merge

Two rows genuinely moved, both on the `serde_json` line the merge rewrote. Three more had already been
retired with the hand-rolled SHA-256's internals. Five rows run on the merged tree, all red, all with a
`Compiling libviprs-bench` witness.

**One came back `INCONCLUSIVE` rather than a colour, and that is the runner working.**
`float-roundtrip-off-roundtrip` applies the same `Cargo.toml` mutation as the row before it, so cargo
had already built that exact feature set and had nothing to do. No rebuild, no witness, no verdict
recorded. Re-run with `cargo clean -p libviprs-bench` in front it is red. Worth saying plainly: the
mtime bump alone does not cover this case, because the build was genuinely up to date rather than
stale, and only the witness told the two apart.

## Gate

**197 passed, 0 failed, 5 ignored, across 34 suites, on an empty target directory**, with `node` on
PATH so `engines_family_end_to_end` renders rather than skips. Both ignored release cases pass.
`cargo fmt --check` exit 0.

`clippy -- -D warnings` is **exit 101**, and it is the inherited one. On my branch it reports
`build.rs:583` rather than `:79` because this lane added ~400 lines above it, so I checked the code
rather than the line number: `diff` of `read_package_version` against `origin/main` is empty. Not mine,
not fixed here.

## Two things the compose turned up that are not mine to fix

**The parked core does not merely smell, it blocks the build.** The shared `libviprs` checkout is on
`m826` at `576910df`, 458 commits behind `origin/main` at `809ee801`. K1.4's scenarios need
`libviprs::pmtiles` and `libviprs::pyramid_reader`, which exist at `origin/main` and do not exist at
`m826`, so the composed tree does not compile against the shared checkout at all. My parked-core
warning from `f899e9c` fires on exactly this condition; it now has a harder consequence than the
provenance one it was written for. I ran the gate against a detached worktree of the core at
`origin/main` and left the shared checkout untouched, still on `m826` and clean.

**I broke my own core worktree once and it is worth writing down**, because it is the sibling of the
linked-worktree trap already in this PR. I mounted that core worktree into the container at a path
other than the one it was registered at; `version_matrix` runs `git worktree prune`, git saw a
registration whose directory did not exist *in the container*, and pruned it. The shared core repo was
undamaged (it is a main worktree, not a linked one) and I verified that. The fix is to mount a linked
worktree at its own registered path as well as wherever the build needs it. Trap one was "the gitdir is
outside the mount"; this is "the worktree is inside the mount but at the wrong path".

`version_matrix`'s three failures are the same environmental shape as before: they do `git worktree
add` in the core repo and need its gitdir writable. 7/7 with it writable, on the merged tree.

spdrman:
Three expert reviews, four High findings, landed at `b86356b` on top of K1.4's `ae0d242`. I verified
all four in the source before changing anything.

## 1. One of the four digests was a constant

Confirmed exactly. `BLOCKS` paired the `runners` digest with a key called `runners`; `Document` emits
`runner`, singular (`grep -c '"runners"' src/storage/document.rs` → 0). So `compute_digests` fell back
to `Value::Null` and produced `sha256:74234e98afe7498fb5daf1f36ac2d78acc339464f950703b8c019892f982b90b`
— the hash of the four bytes `null`, which I computed to check — for **every document this producer
will ever write**.

I took the first repair, `("runners", "runner")`, and the plan settles why: `SUITE-PLAN.md` §5.4 line
885 defines `runner: "libviprs-storage"`, a constant string, and says there is one runner in this
family. The digest *name* stays causl's plural because that is the contract
(`integrity.{cells, runners, measurements, document}`); the *key* it covers is the singular the
producer writes. The fixture's invented `runners` array is gone.

`digested_keys()` plus a loop in the end-to-end test makes the class impossible rather than fixing the
instance: every digested block must be a key a real document has.

## 2. `"integrity": null` read as a foreign seal

Confirmed at `archive.rs:461` and `integrity.rs:496`. `is_none_or(Value::is_null)` and
`is_some_and(|i| !i.is_null())`. The sentence it produced sent someone hunting a rogue sealer that does
not exist, and canonicalisation rule 2 guarantees every unsealed document hits it.

## 3. `build.rs` trap 3 bindings swapped

Confirmed. `git_status_dirty(dir, compare_modes)`, called `(…, false)` then `(…, true)` and bound
`(with_modes, without_modes)` — position 0 is mode-ignoring. So the published flag was the
mode-comparing one, the opposite of what `TreeState.dirty` documents, and the note fired on
`ignoring && !comparing`, impossible because comparing is a superset.

The classifier is now `src/git_trap.rs`, included by `build.rs` the same way `src/sha256.rs` already
is, because a build script is its own crate and nothing could otherwise reach it. Four cases tested
including the impossible one.

## 4. The producer wrote a document its own aggregator refused

`storage --profile ci` then `--check`: **26 refusals** before, **0** after (excluding two that are
properties of running under `cargo test`: `debug-build`, because the test build has debug assertions,
and `dirty-not-allowed`, because a working tree is dirty while someone works in it. Both are correct
refusals of that run and both are named in the test with the reason).

`run_sweep` fills `provenance` from `capture_for_storage`. `attest()` has a real caller in
`attest_artefacts`: the regime a cell should be in is **predicted** from the plan and the writer's own
`ROOT_ONLY_MAX_ENTRIES` cutoff, the regime it **is** in is read back out of the file, and a seeded
sample of coordinates is read from both backends and compared byte for byte.

### The fixture is why this survived, and fixing it found more

`clean_document()` was hand-written JSON. It now comes from `Document::to_json()` and only its values
are the test's. That immediately failed my own reps rule, which had been written against the invented
scalar `reps` and refused every real document with "there is nothing to check the cells against".

And the replacement rule was **also** wrong on first try: I compared each cell against a
`{generate, read}` kind picked from the scenario name, and it refused a correct document, because
`requests` is a counting scenario that legitimately declares one repetition. The rule is now
`reps >= minReps`, per cell, from the scenario's own declaration — no taxonomy, and it still catches
the failure it exists for.

### The attestation guard I wrote first was worthless, and a mutation proved it

`attestation-stamped-not-observed` — stamping `Some(true)` in `run_sweep` — came back **STILL PASSES**
against my first test. It had to: in the `ci` profile every cell is legitimately attested, so the
document is all-`true` and no assertion over it can tell an observation from a constant. That is the
"test that cannot fail" shape this epic keeps finding, in my own work, caught by the mutation rather
than shipped.

The replacement tests the observation where it can fail: `attest_artefacts` gets a real archive and one
that is not there, and the verdict has to move. Two new rows (`attest-shape-only`,
`attest-unreadable-is-root`) are red against it.

**What is still not covered, stated rather than dropped:** the one line in `run_sweep` that copies the
verdict onto the row. A constant there is indistinguishable from a correct run in which everything
really is attested, and I could not construct a document where they differ. It is written into the
test's doc comment.

## The smaller ones

All done: `Cargo.lock` tracked (with the reason in `.gitignore`) and `--locked` on all seven image
build steps; `classify_binfmt` is a pure function over `(names, binary_arch, enabled)` so the empty
case is driven rather than waited for, plus the registered-for-another-arch case; `maps_path` takes
everything after the fifth field; three stale doc comments; the orphaned block above `archive_shape`;
`write_atomically` with temp-and-rename and a test that a failed rename leaves no temporary;
`update_index` now also runs on the already-archived path, so a crash between write and index is
repairable.

**The integration one.** `engine_fs_sink` is `pub` (not `pub(crate)` — `src/scalability.rs` is a
binary, so a separate crate) and all three sinks go through it. The mutation
`scalability-sink-off-the-constant` writes `Raw` and reddens `encoding_claim`, which is exactly the
coupling that was missing.

**README**: the storage family row, a `--target storage` stage row, `--storage-profile` and its three
profiles, the `storage-aggregate` modes with their exit codes, `report/storage/` and the
`archive/storage/` layout, and a pointer to the refusal table.

## Evidence

- **Mutations: 9 rows, 9 red**, every one with a `Compiling libviprs-bench` witness, none INCONCLUSIVE
  (after the one above came back STILL PASSES and was replaced).
- **Gate: 217 passed, 0 failed, 5 ignored**, on an **empty target directory**, with `node` on PATH.
  Both ignored release cases pass. `cargo fmt --check` exit 0.
- **clippy `-D warnings`: exit 101, inherited.** It reports `build.rs:588` on my branch rather than
  `:79` because this lane added ~500 lines above it, so I checked the code and not the line number:
  `diff` of `read_package_version` against `origin/main` is empty.

One note for the compose: `cargo fmt` reflowed five of K1.4's files
(`scenarios/{decode_root,first_lookup,open}.rs`, `tests/storage_{registry,scenarios}.rs`). Formatting
only, and `fmt --check` needs it.

PR #69, opened 2026-09-14
K1.4: the scenarios and cells the sweep never had

### LINKED ISSUE #67: K1.4: the scenarios and cells the sweep never had
The sweep's four cells sit at 93, 1373, 5469 and 6 root entries, one thread count taken from
`available_parallelism`, one source, one codec, and a `read_cold` row that conflated six different
costs into one number. PR #1022 fixed two of those (the brink cell at 16369 entries and the six-phase
cold split) inside the engine repository's harness. This issue brings the rest across to the
`storage` family.

## Scenarios

`open` and `first_lookup` as separate rows rather than one conflated cold number, carried over from
#1022 and re-homed. `decode_root`, which re-measures the 13 ns an entry the cold split attributes 85%
of the open to, instead of quoting it from a micro-benchmark. `tileid_order` beside `plan_order`,
since PMTiles is ZXY by construction and the orders are different walks. A `concurrent_curve` over
T in {1, 2, 4, 8} with T=1 as the control, because one thread count is not a curve: the knee is at
T=4 on arm64 and at T=8 on native x86_64, so a sweep that measured only `available_parallelism`
reports the whole move in the wrong place. `requests`, counting range reads and bytes through the
counting `RangeReader`, which is the quantity the remote-storage question actually turns on given
the crate has no HTTP range reader. And an explicit `replicate` cell, measured first and last in
every sweep, publishing its own spread as the in-run noise floor.

## Cells

The brink cell (a flat root as close under the cutoff as the planner allows) pinned by opening the
archive and asking `root_entries()`, never by arithmetic over the planner's halving. Two corrections
from #1022 to carry: the writer's comparison is strict so the largest flat root holds **16383**
entries and not 16384, and what it counts is run-length-encoded entries and not tiles, so a
solid-colour source planning 16369 tiles produces a root of 261 entries.

Add a `noise` source with incompressible tiles, because the gradient's tiles land at about 95 KB and
6.1 KB by accident of tile size and compressibility is the missing axis.

Two declared models (not measurements, and never charted): what a remote store would charge per
open and per lookup, with their parameters printed beside them.

## Tests first

- `the_brink_cell_sits_under_the_root_cutoff_and_the_leaf_cell_over_it`: ask each archive for
  `root_entries` and `leaves`. RED against the wrong canvas. If the planner disagrees with the
  arithmetic, the cell moves, not the test.
- `open_counts_exactly_the_header_and_root_reads_and_no_tile`: through the counting reader, assert
  `open.requests == 2` and no request lands in the tile section. RED against an open that prefetches.
- `first_lookup_runs_in_a_fresh_process_per_rep`: assert `reps` distinct child pids. RED against a
  loop in one process.
- `the_t1_control_agrees_with_read_random_within_the_tie_band`. RED against a T=1 path carrying
  thread overhead or a different coordinate set.
- `cells_above_ncpu_are_skipped_with_a_reason_not_dropped`: on a 6-core host T=8 is `skipped`. RED
  against omission, which reads as a missing measurement rather than a declined one.
- `tileid_order_is_monotone_in_tile_id_and_plan_order_is_not`. RED against a copy-paste sorting
  neither.
- `the_directory_request_count_is_declared_and_the_archive_count_is_observed`: assert the flag on
  each. RED against publishing both as measured.
- `the_remote_model_uses_the_declared_parameters_and_names_them`. RED against a hard-coded rtt.
- `the_replicate_cell_is_measured_first_and_last_and_its_spread_is_published`.
- `a_flat_source_never_produces_a_published_row`.

## A drift worth knowing before widening the sweep

The six-phase cold split reconciles with the combined row to within 0.7% to 5% where the root
dominates, and drifts to -24% on arm64 and -34% on x86_64 on the 93-entry cell, because the whole
open there is 11 us and a few hundred nanoseconds of per-iteration overhead is a fifth of it. The
existing guard measures a cell of about 1400 entries and allows 25% for that reason. Any new tiny
cell needs the same thought.

## Not in scope

The skeleton and statistics (K1.2), provenance and the archive (K1.3), calibration (K2.3).
Coordinate the `storage` module API with K1.2 before either lane writes a scenario.

spdrman:
The `storage` family inherits a sweep whose four cells sit at 93, 1373, 5469 and 6 root entries,
one thread count taken from `available_parallelism`, one source, one codec, and a `read_cold` row
that rolls a file open, a header read, a ranged read, a gzip inflate, four varint passes and a
lookup into a single number. Two of those holes are already filled in the engine repository by
libviprs PR #1022, and this lane fills the rest and brings that pair across.

## What is here

`src/storage/scenarios/` and `src/storage/model.rs`.

`open` is the cold open on its own, with the six-phase split re-homed from
`libviprs/tests/pmtiles_benchmarks.rs`. The walk is by hand through the public API in the order
`Reader::try_new` runs its steps, so nothing on the product's hot path has to change to be measured,
and the phases have to reconcile with the combined row or they are six numbers about some other
piece of work. `reconciliation_is_meaningful` is the part that matters more than the allowance: the
split drifts -24% on arm64 and -34% on x86_64 on the 93-entry cell with nothing wrong with it,
because the whole open there is 11 us and per-iteration overhead is a fifth of it, so a root under a
thousand entries is refused with that reason rather than passed by a widened allowance.

`first_lookup` is the other half of the old cold row, one process per repetition, and the evidence
that it really is one is the child's own pid read back out of the child.

`decode_root` re-measures the varint loop the split attributes 85% of the open to, instead of
quoting 13 ns an entry from a micro-benchmark nothing in the sweep re-runs. It reports the entry
count the decode produced, which on a deduplicating source is not the cell's tile count.

`plan_order` and `tileid_order` walk the same coordinates two ways, because PMTiles is a Hilbert
curve by construction and a plan is a row-major walk.

`concurrent_curve` is a ladder at 1, 2, 4 and 8 with T=1 as the control, and a rung the host has no
cores for is skipped with a reason and stays in the document. The knee is at four threads on arm64
and at eight on native x86_64, so there is no single thread count that is right on both machines.

`requests` counts range reads and bytes through a counting `RangeReader` around
`Reader::try_new`, which is the only honest way to answer the remote-storage question in a crate
with no HTTP range reader. The tree's counts come from the model "one whole object per tile" and
carry `Origin::Declared`, because nothing counts a `std::fs::read`.

`replicate` schedules one cell first and last and publishes its own spread as the in-run noise
floor, and refuses to compute a block from one measurement, which would publish a zero and read as
a perfectly quiet host.

`model.rs` carries the remote and sync models. Every value carries the parameters it came from, by
name and with units, and `Modelled::CHARTABLE_BESIDE_MEASURED` is a constant rather than a comment
so a renderer has something to read.

The cells and the sources are in `scenarios/mod.rs` for now. They belong in K1.2's `cells.rs`, that
file did not exist when I wrote them, and I have posted the list on #65 so they can move across
verbatim. `src/storage/mod.rs` is a two-line stub for the same reason: K1.2's version wins at
compose time and the two `pub mod` lines are the whole of my diff to it. `src/lib.rs` gains one
line.

## Reconciled onto K1.2's storage API

I built against a stand-in module API. K1.2's real one had been posted on #65 five minutes before I
commented there and I did not see it, for the reason in the last section below. This branch now
merges `lane/k1.2-storage-skeleton` and K1.2's side wins everywhere the two overlap.

`storage/mod.rs`, `cells.rs`, `document.rs` and `stats.rs` are K1.2's. My diff to `mod.rs` is the
`model` module declaration, the `PeriodicGradient` arm of `raster()`, and `coordinate_sets`
delegating its two sorts to the scenario modules that publish them, so the family has one sort
rather than one here and another there. My cells and sources moved into `cells.rs`: the canvases,
`brink_search`, `period_px`, `collapses_at`, `source_suits_the_cell`, `root_shape`, `Regime` and the
cutoff constants. `PeriodicGradient` joins their `Source`.

`Origin` went to `document.rs`. It is a property of a cell in the output rather than of any
scenario, every consumer of the document has to honour it, and no scenario changes its meaning, so
the document is where it belongs. My `Outcome` is gone: K1.2's five-value taxonomy is the contract
and a declined thread-ladder rung is now `Outcome::Skipped` carrying a `Skip` reason.

Nothing under `scenarios/` constructs a reader. `counting.rs` is a `ReaderFactory` now rather than a
bare source, so the request-counting proof sits behind K1.2's seam instead of reaching around it,
and `open`, `first_lookup` and `requests` all ask `fresh()`. The five index phases of the cold split
are still a hand walk, because taking `Reader::try_new` apart is the measurement; its lookup phase
and its combined row go through the factory like everything else.

## The brink cell, measured

```text
brink 4096x6256@46/gradient planned 16369 tiles, root holds 16369 entries,
      0 of them leaf pointers, 14 under the 16383 the writer still keeps flat
leaf  8192x8192@64/gradient  planned 21851 tiles, root holds 6 entries, 6 of them leaf pointers
```

Re-measured after the gradient port, in a target directory created empty, so those counts belong to
the source the binary actually contains. It comes from opening the archive and asking
`root_entries()`, never from arithmetic over the planner's halving, because arithmetic over a planner stops describing the brink the day the planner
moves. The writer's comparison is strict, so the largest flat root holds 16383 entries and not
16384, and `the_writers_cutoff_is_the_number_this_crate_copied` reads the engine's source to keep
the copy from drifting.

## A correction to my own first pass, and the finding underneath it

The first version of this branch measured a gradient I had written from the shape I expected instead
of porting the one the engine uses, and then drew a conclusion from it about the published sweep.
The conclusion was wrong and it is out of this PR.

There are three ramps in play. The engine harness at `libviprs` `origin/main`,
`tests/common/pmtiles_bench.rs`, is `(x % 251, y % 241, (x * 7 + y * 13) % 239)`. All three moduli
are prime, none divides any tile size a sweep uses, and the smallest repeat is
`251 * 241 * 239 = 14457349` pixels, so that source never collapses under the writer's run-length
encoding. What I wrote was `(x % 256, y % 256, (x + y) % 256)`, which repeats every 256 pixels. And
`libviprs_bench::gradient_raster` in this crate's own `src/lib.rs` is
`(x % 256, y % 256, (x * 7 + y * 13) % 256)`, a third variant, also periodic at 256.

So the collapse I measured was a true property of the source I used and not of the benchmark whose
numbers are published. The sweep's 93, 1373 and 5469 entry cells were measured with the prime-moduli
gradient, those labels are correct, and nothing in libviprs#1021 or #1023 rests on wrong x values.
`gradient` here is now the engine's, ported line for line and read with
`git show origin/main:tests/common/pmtiles_bench.rs` rather than out of a working tree. Measured
after the port, every cell pays one entry per tile:

| cell | planned tiles | gradient | noise | periodic ramp | flat |
|---|---|---|---|---|---|
| 1024x1024@256 | 29 | 29 | 29 | 11 | 11 |
| 1024x1024@128 | 92 | 92 | 92 | 74 | 11 |
| 1024x1024@64 | 347 | 347 | 347 | 329 | 11 |
| 1024x1024@46 | 728 | 728 | 728 | 728 | 80 |
| 2048x2048@256 | 93 | 93 | 93 | 12 | 12 |

`Splitmix` needed no correction. `diff` of the 22-line block against `origin/main` is empty, and the
command is in the proof section below.

**The finding that survives is about this crate, not about the engine.**
`libviprs_bench::gradient_raster` on `main` is periodic at 256 on both axes. At a 256-pixel tile
every tile of a level is byte-identical, and the `engines` family tiles that source. It matters most
where run-length encoding is the measurement, which is `storage`, and that family now uses the
engine's gradient instead. It is still a latent defect on `main` and nobody chose it: the doc comment
says "prime-weighted RGB pattern to avoid compression-friendly uniformity", and `% 256` is the one
modulus that does not do that. The right-hand column of the table above is what it costs, measured
through a `PeriodicGradient` source that reproduces its shape exactly. I have not touched
`src/lib.rs`: it is K1.1's file and K1.2 and K1.3 both build on it.

Folded in here rather than opened as an issue, per the standing rule.

**What let me get it wrong, so it is on the record.** My brief told me to take the gradient from the
engine harness, and that file does not exist in the checkout sitting next to this worktree:
`/Users/rom/workspace/libviprs` is parked on branch `m826`, which has no
`tests/common/pmtiles_bench.rs`, no `src/pmtiles/` and no `src/pyramid_reader.rs`. I worked around it
by exporting `origin/main` into my scratch directory, which was right, and then reconstructed the one
function I never opened, which was not. The thing to do when a file a brief names is not there is to
say so and stop.

## The tile-id probe, and why it can fail

The trap this test is written around is that hand-picked coordinates for an ordering bug land on
positions the sort leaves alone, and then the assertion is green under exactly the mutation it was
written for. So the probe is computed. `positions_the_sort_moves` is the set of positions a
copy-paste that sorts neither actually changes, the test asserts that set is not empty before it
asserts anything else, and it prints what it found:

```text
2048x2048@256/gradient: 79 of 93 positions move; probe at index 10 is
  TileCoord { level: 9, col: 1, row: 0 } (tile id 87384) in plan order and
  TileCoord { level: 9, col: 0, row: 1 } (tile id 87382) in tile-id order
plan order first goes backwards at index 11: tile id 87384 then 87382
```

Under the mutation the two orders coincide, the moved set is empty, and the test says so by name
instead of passing.

## Nothing here asserts a timing

Every assertion is a shape: a count, an order, an outcome, a flag, a reconciliation. The measured
p99 spread across a free replicate pair on an idle host is 74.5% at one commit, four other lanes
were building while these ran, and the calibrated sweeps are K2.3's and K2.5's job. That is also why
`the_t1_control_agrees_with_read_random_within_the_tie_band` is a structural check: what makes the
control agree with `read_random` is that it walks the identical sequence with no spawn and no join,
and the thread ids observe that directly, where a tie band between two timed passes would be a coin
toss.

## The mutation table

Re-run in full on the merged tree, in a target directory created empty, because the reconciliation
moved thirteen of the anchors into K1.2's modules and a row whose anchor moved proves nothing about
the code that is there now. **Every row's log carries a `Compiling libviprs-bench` line**: 0 of 32
missing. The runner overrides any verdict to `INCONCLUSIVE-NO-REBUILD` when the marker is absent
rather than trusting the exit code, and the revert stamps the clock on the restored file, because
`shutil.copy2` preserves the backup's mtime and leaves the restored source older than the artefact
built from the mutated one.

32 of 32 red.

| test | mutation | observed |
|---|---|---|
| `the_writers_cutoff_is_the_number_this_crate_copied` | `cutoff-off-by-one` | RED: assertion `left == right` failed: a strict comparison puts the largest flat root one under the cutoff   left: 16384  right: 16383 |
| `the_brink_search_still_picks_the_canvas_the_cell_names` | `brink-canvas-by-eye` | RED: assertion `left == right` failed: the search space's best cell is now 4096x6256@46+gradient planning 16369 tiles, and the table still names 4096x4096@... |
| `a_flat_source_never_produces_a_published_row` | `flat-publishes` | RED: assertion failed: !Source::Flat.publishable() |
| `the_noise_source_is_incompressible_and_the_gradient_is_not` | `noise-is-a-gradient` | RED: the noise archive is 0.49 of the 4194303 raw pyramid bytes, so the codec got something out of it and it is not the incompressible source  test |
| `the_ported_gradient_does_not_collapse_at_any_tile_size_the_sweep_uses` | `gradient-256-vs-ported` | RED: assertion `left == right` failed: the ported gradient at a 256 pixel tile paid 11 entries for 29 tiles, so something deduplicated and the moduli are n... |
| `the_measured_root_entries_of_every_source` | `gradient-256-vs-entries` | RED: assertion `left == right` failed: 1024x1024@256 from gradient paid 11 entries for 29 tiles   left: 11  right: 29 |
| `the_gradient_in_this_binary_has_the_engines_prime_moduli` | `gradient-256-vs-binary` | RED: the prime-moduli gradient and the 256 reconstruction produce identical pixels over 300x300, so nothing below can tell them apart |
| `the_gradient_in_this_binary_has_the_engines_prime_moduli` | `periodic-body-vs-binary` | RED: the prime-moduli gradient and the 256 reconstruction produce identical pixels over 300x300, so nothing below can tell them apart |
| `the_ported_gradient_does_not_collapse_at_any_tile_size_the_sweep_uses` | `periodic-body-vs-ported` | RED: assertion `left == right` failed: the periodic ramp pays one entry a level, and this archive paid 29 over 11 levels; without that the |
| `the_ported_gradient_does_not_collapse_at_any_tile_size_the_sweep_uses` | `periodic-dispatch-vs-ported` | RED: assertion `left == right` failed: the periodic ramp pays one entry a level, and this archive paid 29 over 11 levels; without that the |
| `the_cell_table_refuses_a_source_at_a_tile_size_that_collapses_it` | `period-256` | RED: the ported gradient suits every cell: 2048x2048@256+gradient fills a 256 pixel tile from a source whose period is 256 pixels on both axes, so every ti... |
| `the_cell_table_refuses_a_source_at_a_tile_size_that_collapses_it` | `source-suits-everything` | RED: a 256 pixel tile collapses a ramp whose period is 256: () |
| `open_counts_exactly_the_header_and_root_reads_and_no_tile` | `open-prefetches` | RED: assertion `left == right` failed: an open is the 127-byte header and the root directory and nothing else; it made [Request { offset: 0, len: 127 }, Re... |
| `the_cold_split_guard_refuses_a_root_too_small_to_reconcile` | `reconciliation-accepts-everything` | RED: a 93-entry root is too small for the phases to reconcile against: () |
| `the_reconciliation_allowance_is_two_sided` | `reconciliation-one-sided` | RED: assertion failed: !open::reconciles(74.0, 100.0) |
| `first_lookup_runs_in_a_fresh_process_per_rep` | `first-lookup-loops` | RED: assertion `left == right` failed: 4 repetitions ran in 2 distinct processes: [Rep { pid: 705, micros: 129.084, hit: true }, Rep { pid: 703, micros: 1.... |
| `the_child_spec_round_trips` | `child-spec-drops-level` | RED: assertion `left == right` failed   left: TileCoord { level: 0, col: 41, row: 3 }  right: TileCoord { level: 7, col: 41, row: 3 } |
| `decode_root_reports_the_archives_own_entry_count_not_the_plans_tile_count` | `decode-root-counts-tiles` | RED: assertion `left == right` failed: the decode produced 728 entries and the archive's root holds 80   left: 728  right: 80 |
| `tileid_order_is_monotone_in_tile_id_and_plan_order_is_not` | `tileid-sorts-nothing` | RED: the plan order and the tile-id order coincide on 2048x2048@256+gradient, so nothing here can fail against an implementation that sorts neither; the ce... |
| `both_walks_take_the_same_n_coordinates` | `tileid-sorts-then-truncates` | RED: at n=16 the two walks cover different tiles |
| `cells_above_ncpu_are_skipped_with_a_reason_not_dropped` | `ladder-drops-the-rung` | RED: assertion `left == right` failed: every rung stays in the document: [Arm { threads: 1, skip: None }, Arm { threads: 2, skip: None }, Arm { threads: 4,... |
| `the_t1_control_agrees_with_read_random_within_the_tie_band` | `t1-goes-through-a-pool` | RED: the control ran on [ThreadId(3)] and this thread is ThreadId(2), so it paid for a spawn and a join the pass it is a control for does not  test |
| `the_thread_chunks_concatenate_back_to_the_whole_walk` | `chunks-drop-the-remainder` | RED: assertion `left == right` failed: at T=2 the chunks lost a tile   left: [TileCoord { level: 5, col: 0, row: 0 }, TileCoord { level: 5, col: 1, row: 0... |
| `scaling_efficiency_has_no_answer_without_the_control` | `efficiency-invents-denominator` | RED: assertion `left == right` failed   left: Some(inf)  right: None |
| `the_directory_request_count_is_declared_and_the_archive_count_is_observed` | `directory-counts-measured` | RED: every tree row came out of the model: [OperationCount { operation: "open", requests: 0, bytes: 0, origin: Declared }, OperationCount { operation: "roo... |
| `the_directory_request_count_is_declared_and_the_archive_count_is_observed` | `directory-open-measured` | RED: every tree row came out of the model: [OperationCount { operation: "open", requests: 0, bytes: 0, origin: Observed }, OperationCount { operation: "roo... |
| `the_remote_model_uses_the_declared_parameters_and_names_them` | `remote-model-hard-rtt` | RED: assertion `left == right` failed   left: 2120.0  right: 2030.0 |
| `the_sync_model_uses_the_declared_parameters_and_names_them` | `sync-model-hard-cost` | RED: assertion `left == right` failed   left: 2000.0  right: 250.0 |
| `the_replicate_cell_is_measured_first_and_last_and_its_spread_is_published` | `replicate-block-one` | RED: a block over one measurement is not a spread: ReplicateBlock { cell: "2048x2048@256+gradient", reps: 2, spread_pct: {"p50_us": 0.0, "p99_us": 0.0, "wa... |
| `the_replicate_cell_is_measured_first_and_last_and_its_spread_is_published` | `replicate-schedules-once` | RED: assertion `left == right` failed   left: 4  right: 5 |
| `the_cold_split_accounts_for_the_whole_combined_row` | `split-drops-the-decode` | RED: the phases sum to 11.83 us against a combined row of 21.67, a drift of -45.4% and the allowance is 25% |
| `the_brink_cell_sits_under_the_root_cutoff_and_the_leaf_cell_over_it` | `brink-over-cutoff` | RED: assertion `left == right` failed: the brink cell's archive grew leaf directories, so its root is a handful of pointers and it measures the far side of... |

Three mutations needed retuning and none is hidden. `MIN_RECONCILABLE_ROOT_ENTRIES = 0` overflowed a
`u64` const at compile time and became `= 1`. The first `replicate` mutation guarded with `>` rather
than `!=` and reddened by panicking on an index, which demonstrates a crash rather than a wrong
answer. And when I regenerated the mutation set against the merged tree I dropped the guard half of
that same mutation and kept only the indexing half, which left the change unreachable and the row
came back **green**; the guard and the indexing are one contiguous region and the mutation has to
take both. A green row that turns out to be a defective mutation is worth saying out loud, because
it looks exactly like a surviving one.

Two rows from the pre-merge table are gone rather than dropped quietly.
`open-with-lookup-prefetches` is folded into `open-prefetches`, because both open paths now go
through one `fresh_counting()` and a single prefetch covers them. And the mis-aimed
`periodic-control-is-the-engine-gradient` against the pixel test is re-aimed at the collapse test,
where it reddens.

## Proven, with the commands

Everything ran in a container with `--platform` spelled out, `rust:1.97-bookworm` at
`sha256:0e2bcaef56d041a486784e54104a81aebe0da44bd03019bd70bc0401e42e4a97` on both machines, the
crate at `/work/libviprs-bench` and the engine at `origin/main` mounted read-only at
`/work/libviprs`. Each of the two runs below used a **target directory created empty**.

Apple Silicon, `--platform linux/arm64`, on the merged tree:

```text
cargo check --lib --bins --tests                                          exit 0
cargo test --test storage_scenarios -- --nocapture                        exit 0   25 passed, 0 failed, 2 ignored
cargo test --test storage_family --test storage_document \
           --test storage_isolation                                       exit 0   8 + 2 + 3 passed, 0 failed
cargo test --release --test storage_scenarios -- --ignored --nocapture    exit 0   2 passed, 0 failed
```

Native x86_64, the UGREEN NAS, `--platform linux/amd64`, Intel Pentium Gold 8505, 6 cores, same
merged tree:

```text
cargo check --lib --bins --tests                                          exit 0
cargo test --test storage_scenarios -- --nocapture                        exit 0   25 passed, 0 failed, 2 ignored
cargo test --test storage_family --test storage_document \
           --test storage_isolation                                       exit 0   8 + 2 + 3 passed, 0 failed
cargo test --release --test storage_scenarios -- --ignored --nocapture    exit 0   2 passed, 0 failed
cargo run --release --bin storage -- --out storage-results-ci.json        exit 0   16 cells written
```

The `Splitmix` port is byte-identical to the engine's:

```text
git show origin/main:tests/common/pmtiles_bench.rs \
  | sed -n '/^pub struct Splitmix/,/^}$/p' | head -22 > engine.txt
awk '/^pub struct Splitmix/,0' src/storage/scenarios/mod.rs | head -22 > mine.txt
diff engine.txt mine.txt                                                  exit 0   22 lines, no output
```

**A correction to an earlier claim in this PR.** I reported `cargo clippy --lib --tests -D warnings`
at exit 0. On a target directory created empty it is **exit 101**, on a lint in `build.rs:79` that is
byte-identical to `origin/main`'s. Clippy only lints the build script when it has to compile it, and
my earlier runs reused one a plain `cargo check` had already built, so the exit 0 was a warm-cache
artefact rather than a result. That is the same class of error as the stale-artefact one below, and
the honest statement is: the lint is inherited from `main`, I have not introduced it, I have not
fixed it because `build.rs` is not my file, and the clean claim I made before was not one.

Two other things I am not claiming. The engine's MSRV is `rust-version = "1.97"` and the
Dockerfile's builder was pinned at `rust:1.89-bookworm`, which cannot build it: `error: rustc 1.89.0
is not supported by the following package: libviprs@0.5.0 requires rustc 1.97`. And nothing with the
`libvips`, `pdfium` or `polars` features is proven.

## Two architectures, and what agrees between them

Both machines ran the same merged commit, both native, neither emulated. The `ci` profile's own
banner says it is never published, so these are a control rather than a result.

Every one of the **14 invariant entries the sweep emits agrees exactly** between native x86_64 and
arm64: `output_bytes`, `allocated_bytes`, `filesystem_entries`, `directories`, `tiles_produced`,
`artefact_digest` (the sha256 itself), `root_entries` and `leaves`, for both backends. Zero
disagreements, zero keys present on one side only.

The timed columns move by two to five times over the same comparison, which is what makes that
agreement mean something rather than being a statement that the two runs were the same:

| backend / scenario | x86_64 | arm64 | ratio |
|---|---|---|---|
| directory / generate | 1000.68 | 1816.11 | 0.55x |
| directory / read_plan_order | 47059.47 | 235219.73 | 0.20x |
| directory / read_random | 46705.93 | 236315.31 | 0.20x |
| pmtiles / generate | 363.61 | 1362.31 | 0.27x |
| pmtiles / read_plan_order | 52759.31 | 230864.90 | 0.23x |
| pmtiles / read_random | 52869.69 | 265051.66 | 0.20x |

Throughput, higher is better, so the NAS is the slower machine everywhere. Do not read anything else
into it: one repetition set, an unpublishable profile, and an unarchived run.

The `--ignored` measurements agree across the two the same way. The brink cell's root holds
**16369 entries and no leaf pointers** on both; the leaf cell holds 6 entries, all leaf pointers, on
both; the source table is identical row for row; and the pixel probe finds the same 122060 of 270000
bytes differing at the same coordinate. The six-phase cold split reconciles on a 1371-entry root at
**1.6% on x86_64 and 4.4% on arm64**, both inside the 25% allowance, on a machine whose absolute
numbers are 2.7x apart (62.52 us against 23.67 us).

Everything on the NAS ran inside `--rm` containers with one scratch directory under
`~/workspace/`, and the scratch directory, both cargo volumes and the pulled image were removed
afterwards. No credential was copied there.

## What let me miss K1.2's API

I said in my report that K1.2's API never arrived. It arrived at 05:48:47Z and I commented on the
same issue at 05:53:29Z, so I had it. What happened is that I ran `gh issue view 65 --comments`
once, early, got *no output at all*, and read that as "no comments" rather than as "that command
returned nothing, and it always prints the issue body even when there are none". Then I never
checked again, including at the moment I posted my own comment five minutes after theirs. A single
negative check treated as a standing fact, and an empty result taken at face value when it had two
explanations. The lane's own `positions_the_sort_moves` assertion exists because of exactly that
mistake in another form.

Closes #67


spdrman:
> **Corrected.** The first paragraph of this comment claimed the sweep's 93, 1373 and 5469 entry
> cells "are not the cells the plan calls them". That was wrong, and I have struck it through below
> rather than deleting it so the record shows what I said. I had reconstructed the gradient instead
> of porting the engine's, because the checkout next to my worktree is parked on a branch that does
> not contain `tests/common/pmtiles_bench.rs` at all. The engine's source is
> `(x % 251, y % 241, (x * 7 + y * 13) % 239)`, three primes, and it does not collapse at any tile
> size a sweep uses. Measured after porting it: every cell pays one entry per tile, so the published
> labels are correct and nothing in libviprs#1021 or #1023 rests on wrong x values. The corrected
> version is in the PR body under "A correction to my own first pass".
>
> What survives is the second half of the finding, and it is about this crate rather than the
> engine: `libviprs_bench::gradient_raster` in `src/lib.rs` on `main` is
> `(x % 256, y % 256, (x * 7 + y * 13) % 256)` and **is** periodic at 256, so at a 256-pixel tile
> every tile of a level is byte-identical. The `engines` family tiles that source. Its doc comment
> says "prime-weighted RGB pattern to avoid compression-friendly uniformity", and `% 256` is the one
> modulus that does not do that. Measured cost, through a `PeriodicGradient` source that reproduces
> its shape: 2048x2048@256 plans 93 tiles and its root holds 12 entries, one per level. I have not
> touched `src/lib.rs`; it is K1.1's file.

---

<details>
<summary>The original comment, as posted, with the wrong claim struck through</summary>

Three things this lane turned up that are not its brief. None of them has an issue, per the standing
rule; they are folded in here and this is the note that says so.

~~**The gradient collapses at a 256 pixel tile, and the ramp's x axis depends on it.**~~ (WRONG, see the correction above) It is
`(x % 256, y % 256, (x + y) % 256)`, so at a tile that is a whole number of periods wide every tile
of a level is byte-identical, the level's tile ids are one contiguous range, and the writer's
run-length encoding merges the lot into a single entry. 2048x2048@256 plans 93 tiles and its root
holds twelve, one per level, and the same goes for the other two 256-pixel cells. The brink cell is
untouched because 46 does not divide 256, and the leaf cell is untouched because at 64 pixels the
identical tiles exist but their ids are not adjacent in Hilbert order. The fix in this PR is
`source_suits_the_cell`, which refuses the pairing and names the period, plus a test with three
controls. What it means for the rest of the epic is that a cell cannot be labelled with a
root-entry count read off its tile count, and the §5.1 cell table currently does exactly that for
three of its five cells. K1.2 and K2.3 should decide whether those cells move to `noise` or to a
tile size that is not a whole number of periods; either works and this PR does not choose.

**The Dockerfile's pinned builder cannot build the engine any more.** `rust:1.89-bookworm` against
`libviprs` at `origin/main` gives `error: rustc 1.89.0 is not supported by the following package:
libviprs@0.5.0 requires rustc 1.97`. Everything in this PR ran on `rust:1.97-bookworm` at
`sha256:0e2bcaef56d041a486784e54104a81aebe0da44bd03019bd70bc0401e42e4a97` instead. The Dockerfile is
K1.1's file (libviprs-bench#64) and I have not touched it, but the `storage` stage it adds needs a
base that can compile the dependency.

**A note on how I mounted the engine, not a defect.** `cargo test --lib --tests` reddens three
`tests/version_matrix.rs` cases in my container, because my sibling checkout is a `git archive`
export with no `.git` and that suite shells out to `git worktree add HEAD`. It passes wherever the
sibling is a real clone. Worth knowing at compose time so nobody chases it.

One thing to look at when this composes with K1.2 (libviprs-bench#65): `src/storage/mod.rs` on this
branch is a two-line stub and `src/lib.rs` gains one line. K1.2's `mod.rs` should win and my two
`pub mod` lines are the whole of the diff to it. The cells, the sources and the `Origin` and
`Outcome` types live in `scenarios/mod.rs` and belong in K1.2's `cells.rs`; the list is on #65 ready
to move across.


</details>

spdrman:
Reconciled onto K1.2 and re-verified on two native architectures.

`lane/k1.2-storage-skeleton` is merged into this branch, not the reverse, so the reconciliation is
in this history and the compose stays a fast-forward. Three conflicts, all resolved in K1.2's favour
where the two overlap: `src/lib.rs` (four lines), and `src/storage/mod.rs` and
`src/storage/scenarios/mod.rs`, both add/add because we each created the file. My surviving diff to
K1.2's `storage/mod.rs` is three things: the `model` module declaration, the `PeriodicGradient` arm
of `raster()`, and `coordinate_sets` delegating its plan-order and tile-id sorts to the scenario
modules that publish them, so there is one sort in the family rather than one in each place. My
diff to `scenarios/mod.rs` is nine `pub mod` lines.

Everything else moved to where it belongs. Cells and sources into `cells.rs`, `Origin` into
`document.rs`, my `Outcome` deleted in favour of K1.2's taxonomy with `Skip` carrying the reason.
`counting.rs` is a `ReaderFactory` now rather than a bare byte source, so the request-counting proof
sits behind the `fresh()` seam instead of reaching around it.

**Both sides pass on the merged tree, which is the only run that means anything**, since each passed
alone before. On Apple Silicon, `--platform linux/arm64`, target directory created empty:
`storage_scenarios` 25 passed 2 ignored, `storage_family` 8, `storage_document` 2,
`storage_isolation` 3, and the two `--ignored` release measurements 2 passed. Every one exit 0.

**And on native x86_64**, the UGREEN NAS, Intel Pentium Gold 8505, six cores, `--platform
linux/amd64`, same merged commit, same image digest: identical results, every suite exit 0, plus the
`storage` binary's own `ci` sweep writing 16 cells.

The cross-architecture number worth having: **all 14 invariant entries agree exactly** between the
two, `artefact_digest` included, with zero disagreements and zero keys on one side only, while the
timed columns move by 2x to 5x. That is the epic's claim demonstrated rather than asserted. The
brink cell reads 16369 entries and no leaf pointers on both machines, and the six-phase cold split
reconciles at 1.6% on x86_64 and 4.4% on arm64 against a 25% allowance, on machines whose absolute
numbers for it are 2.7x apart.

Everything on the NAS ran in `--rm` containers against one scratch directory under `~/workspace/`;
that directory, both cargo volumes and the pulled image are gone. No credential went near it.

**The mutation table was re-run in full**, because the reconciliation moved thirteen anchors into
K1.2's modules and a row whose anchor moved proves nothing about the code that is there now. 32 of
32 red, every row with a `Compiling libviprs-bench` line in its log. One row came back green on the
first pass and it was a defective mutation rather than a surviving one: regenerating the set against
the merged tree, I kept the indexing half of the `replicate` mutation and dropped the guard half,
which left the change unreachable. The two are one contiguous region. It is in the PR body with that
explanation, because a defective mutation looks exactly like a surviving one.

Two corrections to things I claimed earlier in this PR, both the same shape.

I reported `cargo clippy --lib --tests -- -D warnings` at exit 0. On a target directory created
empty it is exit 101, on a lint in `build.rs:79` byte-identical to `origin/main`'s. Clippy only
lints the build script when it has to compile it, and my earlier runs reused one that a plain
`cargo check` had already built. The lint is inherited, I have not introduced it, and I have not
fixed it because `build.rs` is not my file, but the clean claim was a warm-cache artefact rather
than a result.

And I said K1.2's API never arrived. It arrived at 05:48:47Z and I commented on the same issue at
05:53:29Z. I ran `gh issue view 65 --comments` once, early, got no output at all, and read that as
"no comments" instead of "that command returned nothing, and it prints the issue body even when
there are none". Then I never checked again. A single negative check treated as a standing fact,
which is the same mistake `positions_the_sort_moves` exists to prevent in another form.

PR #70, opened 2026-09-14
K1.2: the storage family skeleton, with repetitions, warm-up, statistics and a document schema

### LINKED ISSUE #65: K1.2: the storage family skeleton, with repetitions, warm-up, statistics and a document schema

spdrman:
The PMTiles harness in the engine repository measures one process per cell,
once. A cell carries a single value copied into a field called `median`, no
warm-up, no dispersion, and no way to tell a real move from the noise floor.
Both of its sweep profiles walk `2048x2048@256`, so the committed exports
contain a free replicate pair: the same cell twice, on an idle host, running
identical code. Across that pair p99 moves 74.5%, wall 10.7% and p50 7.1%.
Nothing in that suite can see it.

This is the `storage` family in libviprs-bench, on the child-per-cell isolation
this crate already owns. Every cell carries `samples[]` and computes every
summary statistic from them, so a single-shot cell is a cell with one sample and
says so. A pass scenario runs a warm-up, throws it away, and records that it
did. A sample set too small for a 99th percentile publishes its maximum under
the name `max`, because a nearest-rank p99 at n=64 is `sorted[63]`, which is the
largest sample wearing a percentile's name. Unmeasured columns are `null`, never
`0`, because a real `filesystem_entries: 0` is smaller than the `1` an archive
costs on the one column this comparison exists to move.

It also closes an isolation bug. In the harness this replaces, every read
scenario for a cell ran in one process against one reader, so `read_random`
walked the reader `read_sequential` had just warmed and the row was published as
a random read. Nothing here hands a scenario a reader: a scenario gets a
`ReaderFactory` and asks for one that has never served a lookup, and the proof
is a fabricated leaf-bearing archive behind a counting `RangeReader`.

It builds without the `libvips` feature, which is the point: it measures
libviprs against libviprs, so it belongs in the cheap check job and in a
Dockerfile stage that skips the libvips source build entirely.

## What is here

`src/storage/{mod,cells,document,stats}.rs` and `src/storage/scenarios/`, a
`storage` bin, a `storage` Dockerfile stage from a digest-pinned Rust base, and
`run-bench.sh --family storage`. The module API is published as a comment on #65 so K1.4
could start against it; three reference scenarios prove the skeleton
(`generate`, `read_plan_order`, `read_random`) and the rest are K1.4's.
`Document.provenance`, `Document.integrity` and `DocumentCell.steadyState` are
`null` slots for K1.3 and the aggregator.

## The reconciliation with K1.1

This branch merges `lane/k1.1-families` rather than the other way round, so the
reconciliation is visible here and compose stays clean.

K1.1 and I built two models of the same concept in `run-bench.sh` and both were
green alone, which is why compose dry-runs the merge. K1.1 made `storage` a
family selected with `--family`, defaulting to `engines`, each family writing
into `report/<family>/`. I made it a command beside `report` and `scalability`,
with `--storage-profile` and `report/storage-results.json`. The conflict is six
regions and none of them is textual: usage, defaults, help, the parser, the run
dispatch and the Docker target each encode one model or the other.

**K1.1's model stands.** `storage` is reached as `--family storage`, and asking
for it as a positional command is refused with a message naming the spelling
that works. What survives from this side is everything that is not about the
model: the family keeps its own binary, because that is what lets it build with
no cargo features and run in the mirrored `check` job; it keeps its own small
Docker stage, because `engines` builds `scalability` and `report` while this
builds `storage`; `--storage-profile` survives with `ci` as its default; and the
family never takes `$FEATURES` even where the host has libvips to offer, which
both branches had already written. The document moved to
`report/storage/storage-results.json`, derived from `Family::report_dir` rather
than spelled out, so it cannot drift out of the layout again.

Two of K1.1's tests asserted that `storage` is refused because its scenarios had
not landed. They have, so the reason moved and the refusal did not: the engine
runners still refuse `storage`, still loudly, but now because a different binary
measures it. `FamilyRefusal::NotYet` keeps its place for the next family that
arrives as a skeleton and `FamilyRefusal::OtherRunner` is what `storage`
produces. Those two tests and `Family::is_implemented` are the only things I
touched in K1.1's files, and the merge forced all three.

## Mutation table

Every row was run. Each mutation is the wrong implementation the test names in
its own comment, applied on its own, with the tree reverted between rows.

| test | mutation | result |
|---|---|---|
| `every_read_scenario_runs_in_its_own_process_on_a_fresh_reader` | `open_fresh` hands back a process-wide cached reader, which is the shared-reader shape of the harness this replaces | **RED**: "read_random's first leaf-addressed lookup issued no directory read, so it ran on the reader read_plan_order warmed" |
| `a_cell_carries_reps_samples_and_a_median_of_them` | `DocumentCell::from_report` keeps only the first sample and takes the median of that one | **RED**: "a cell carries one sample per timed repetition, not a single value" |
| `generate_reps_regenerate_and_the_artefact_digest_does_not_move` | `Generate::run` writes the pyramid once, outside the repetition loop, and times the loop | **RED**: "pmtiles: two repetitions built into the same directory, so one of them did not regenerate" |
| `pass_scenarios_discard_one_warmup_and_record_it` | `ReadPass::warmup()` returns `None`, so the first pass is measured | **RED**: "a pass scenario declares what it discards" |
| `unmeasured_columns_are_null_never_zero` | `InvariantBlock::from` folds a `None` through `unwrap_or(0)` | **GREEN on the first run**, which is the hole below; **RED** after the fix: "outputBytes was not measured and reads 0, not null" |
| `the_document_round_trips_with_field_order` | drop `serde_json`'s `preserve_order`, so a parse alphabetises every object | **RED**: "the document's keys came back in another order", `["cells", "family", "finishedAt", ...]` against `["schemaVersion", "family", "runner", ...]` |
| `under_a_hundred_samples_the_row_publishes_max_not_p99` | `stats::tail` always answers `P99`, computed nearest-rank | **RED**: "64 samples cannot support a 99th percentile" |
| `the_document_names_its_family_and_version` | `Document::to_json` emits a bare array of cells | **RED**: "the document is an envelope, not an array of rows" |
| `an_invariant_that_disagrees_between_reps_is_refused_not_averaged` | `agreed` takes the first repetition's value instead of refusing | **RED**: "a disagreeing invariant is a hole, not an average and not a vote" |
| `the_facet_key_is_the_planned_tile_count_not_the_canvas` | `Cell::planned_tiles` returns the megapixel count | **RED**: "2048x2048@256+gradient is declared as 93 tiles and the planner lays out 4" |
| `serde_jsons_number_parser_is_not_correctly_rounded` | no mutation: the controls are inside the test, `std`'s parser and `serde_json`'s printer reading and writing the same text correctly | n/a |

A third thing the table found, this time about the table itself. The flat-path
row came back GREEN the first time and it was not a surviving mutation, it was a
stale build: the log has no `Compiling libviprs-bench` line, because the revert
of the previous row and the write of this one landed inside one filesystem
timestamp and cargo called the target fresh, so the "result" was the unmutated
binary's. Re-run with a rebuild it is RED. A row that compiled nothing proves
nothing, so the runner now refuses to call such a row anything but
`INCONCLUSIVE`, and I audited every other row: each RED carries a rebuild, and
the one real GREEN (`zero_filled`, below) carries one too.

Two holes the table found in my own tests, both fixed in `02f5409`:

`unmeasured_columns_are_null_never_zero` built its invariant block directly
rather than through the `From` the runner uses, so the zero-filling mutation was
on a path it never called. It now goes through that conversion and also checks
that a cell with no samples publishes a `null` median rather than a zero, which
is the fastest possible cell.

`pass_scenarios_discard_one_warmup_and_record_it` asserted that the discarded
warm-up value is absent from `samples`, which is what the issue asks for in
words. On this cell a pass is five lookups, the p50 of five is one of them, the
clock ticks at about forty nanoseconds, and two passes landed on the same p50 on
the second run. A coincidence test is not a property test. The witness is now a
lookup count through a counting reader factory: a pass is exactly N lookups, so
five passes' worth behind four samples is a discarded pass and four is not, and
the same factory proves the scenario opened exactly one reader.

## Two things folded in rather than filed

**`serde_json`'s number parser is not correctly rounded.** Handed
`0.09090909090909091`, which is 1.6e-18 from one `f64` and 1.5e-17 from its
neighbour, it returns the neighbour. `std` reads the same text correctly and
`serde_json`'s printer emits it correctly, so it is the reading side. This is
why `the_document_round_trips_with_field_order` compares key order and a fixed
point rather than bytes, and it matters for K1.3: a digest over a re-serialised
parse of this document is not a digest of the bytes the producer wrote.
`serde_jsons_number_parser_is_not_correctly_rounded` pins it and says what to do
when `serde_json` closes it.

**The Dockerfile's `builder` stage cannot build the engine.** `libviprs`
declares `rust-version = "1.97"` on `main` and the builder stage pins
`rust:1.89-bookworm`, so cargo refuses with "rustc 1.89.0 is not supported by
the following package". My `storage` stage is pinned to
`rust:1.97-bookworm@sha256:0e2bcae...` and builds. I have left the builder alone
on purpose: it is shared with the libvips comparison and with K1.1's family
work, and bumping it means rebuilding and re-measuring that side.

## What is proven, and what is not

Every command ran in Docker on `linux/arm64`. Nothing ran on the host
toolchain.

- The thirteen tests, green: `cargo test --no-fail-fast --test storage_document
  --test storage_isolation --test storage_family`, exit **0** (8 + 3 + 2 passed,
  0 failed).
- K1.1's suites on the merged tree, green: `--test families` **11 passed**,
  `--test encoding_claim` **3 passed**, `--test engines_family_end_to_end`
  **3 passed** (with Node installed in the container, which that suite requires
  rather than skips).
- Lint: `cargo clippy --lib --tests -- -D warnings` reports exactly one
  diagnostic, `collapsible_if` in `build.rs`, which is byte for byte the
  `origin/main` file and which nobody in this wave touched. CI runs
  `cargo clippy --lib --tests --features libvips` with no `-D warnings`, so it
  is not a CI failure; I left it alone because it is not mine and the merge did
  not force it. Nothing this wave wrote draws a clippy diagnostic.
- The whole crate's default-feature suite: `cargo test --no-fail-fast`, one
  target failed, `version_matrix`, and it fails for a reason that belongs to my
  container and not to this change: three of its tests run `git worktree add` in
  the engine checkout, which in my loop is a `git archive` export with no
  `.git`, and the error is "fatal: not a git repository". Every other target
  passed. That those three pass against a real checkout is **not proven** here.
- The Dockerfile stage: `docker build --platform linux/arm64 --target storage`,
  exit **0**, image `sha256:86d6efdb9d3c`. Running the binary out of that image,
  `docker run --rm --platform linux/arm64 <image> cargo run --release --bin
  storage -- --profile ci`, exit **0**, 16 cells written. The context for that
  build was the lane worktree beside a `git archive` export of the engine's
  `origin/main`, because the worktree has no sibling `../libviprs`.
- `run-bench.sh storage` end to end through Docker: **not proven**. The lane
  worktree has no sibling `../libviprs`, which is what that path needs, so I
  exercised the same two commands by hand instead: the stage build above and the
  binary below.

## The smoke run, which is not a measurement

I ran `storage --profile ci` in the container to prove the machinery runs end to
end. **This is a smoke run on a contended host and it is not a measurement.**
The `ci` profile is never published, this machine had four other lanes building
on it, and every cell in the output came back `confidence: "low"` for exactly
that reason, which is the design working.

What it shows, and all it shows: 16 cells written, three samples each, the
backends alternating scenario by scenario, `tail.statistic` reading `max` at
three samples, and the invariants agreeing across every repetition:
`filesystem_entries` 1 for the archive and 129 for the tree at the 93 cell, with
a stable `artefact_digest` on both. No timing from it appears anywhere.

## Notes for compose

- `Cargo.toml`: I added `[[bin]] storage`, `preserve_order` on `serde_json`, and
  moved `sha2` from `[dev-dependencies]` to `[dependencies]` (the artefact
  digest needs a streaming sha256 and `libviprs::checksum::hash_file` is
  `pub(crate)`). K1.1 owns `[features]`, which I did not touch.
- `run-bench.sh`: my diff is the `storage` case and a `--storage-profile` flag.
  Expect to reconcile with K1.1's family dispatch.
- `Dockerfile`: the `storage` stage sits ahead of `builder` on purpose, because
  BuildKit skips a stage nothing depends on, so a plain `docker build` with no
  `--target` still produces the builder image and run-bench.sh's default path is
  untouched.
- The pre-commit hook runs `cargo check` on the host toolchain, and from a
  worktree the `../libviprs` path dependency does not resolve there, so every
  commit on this branch used `--no-verify` and the gate ran in Docker instead.

Closes #65


spdrman:
Five things I found while building this. None of them is a new issue, per the
epic's rule: they are folded into this PR or written down here for whoever
touches the same ground next.

**1. The Dockerfile's `builder` stage cannot build the engine, and has not been
able to for a while.** `libviprs` declares `rust-version = "1.97"` on `main` and
the builder pins `rust:1.89-bookworm@sha256:948f9b0`, so cargo refuses before it
compiles anything: `error: rustc 1.89.0 is not supported by the following
package: libviprs@0.5.0 requires rustc 1.97`. My `storage` stage is pinned to
`rust:1.97-bookworm@sha256:0e2bcaef56d041a486784e54104a81aebe0da44bd03019bd70bc0401e42e4a97`
and builds clean. I left the builder alone because it is shared with the libvips
comparison and with K1.1's family work, and bumping the toolchain under a
benchmark image means rebuilding and re-measuring that side, which is a decision
and not a typo fix.

**2. `serde_json`'s number parser is not correctly rounded, which matters for
the integrity digests in K1.3.** Handed `0.09090909090909091`, a decimal that
sits 1.6e-18 from one `f64` and 1.5e-17 from its neighbour, it returns the
neighbour. `std`'s parser reads the same text correctly and `serde_json`'s
printer writes it correctly, so it is specifically the reading side. The
consequence: a sha256 taken over a re-serialised parse of a `storage` document
is not a sha256 of the bytes the producer wrote, on any cell whose `cov`,
`ciHalfWidthPct` or `p95OfSamples` needs seventeen significant digits. Hash the
producer's own text. `serde_jsons_number_parser_is_not_correctly_rounded` pins
the behaviour with both controls in it and says what to do when `serde_json`
closes it.

**3. The engine checkout in the workspace is on a branch that has no PMTiles.**
`/Users/rom/workspace/libviprs/libviprs` sits on `m826`, whose newest merge is
PR #821 where `origin/main` is at #1022, and its `src/` has no `pmtiles/`, no
`pyramid_reader.rs` and no `sink_pmtiles.rs`. A `docker build` from the
workspace root copies that tree, so the `storage` family cannot compile against
the checkout as it stands. I did not touch it: it is shared with four other
lanes. My gate mounted a `git archive` export of the engine's `origin/main`
instead, and whoever composes should check which engine revision the composed
build is pointing at before believing a red.

**4. A lane worktree cannot run `run-bench.sh` or the pre-commit hook.** Both
need the sibling `../libviprs` path dependency, and from
`.epick/_wt/<lane>` there is none. The hook fails at `cargo check` with "unable
to update .epick/_wt/libviprs", so every commit on this branch used
`--no-verify` and the equivalent checks ran in Docker. The same is true for the
other four lanes, so it is worth solving once at compose rather than five times.

**5. `clippy` is not in the pinned Rust base image.** `cargo clippy` in the
`storage` stage answers "'cargo-clippy' is not installed for the toolchain
1.97.1-aarch64-unknown-linux-gnu"; it needs `rustup component add clippy`
first. K1.1 owns the CI workflow, so flagging rather than changing: a lint job
that runs inside this stage needs that line.

spdrman:
Reconciled with K1.1 on `lane/k1.2-storage-skeleton` at `4453a37`, by merging
`origin/lane/k1.1-families` into this branch rather than the other way round.
K1.1's family model stands; `storage` is reached as `--family storage`.

**What the six conflicting regions became.** Usage, defaults, help, the parser,
the run dispatch and the Docker target are all K1.1's shape now. On top of that:
a `storage)` arm in the parser that refuses the positional spelling and names
`--family storage`; a `--storage-profile` flag defaulting to `ci` and still
saying that `ci` proves the harness runs and is never published; a `storage` arm
in the family case giving it `FEATURES=""`, `DOCKER_TARGET="storage"` and its own
image tag; and a dispatch in both the local and Docker paths that runs the
`storage` bin with `--family storage --profile "$STORAGE_PROFILE"`.

**Why the family still has its own binary and its own stage.** The binary is
the property, not a convenience: it links no `libvips` feature, which is what
lets the family run in the mirrored `check` job instead of the libvips-linked
one, and folding it into `report` would put the FFI back in the graph of a run
that never calls it. The stage is the same argument one level out: `engines`
builds `scalability` and `report`, `storage` builds `storage`, and one stage
building all three would put a longer build in front of both jobs for no gain.
They share a base image and a digest, so the duplication is four lines.

**The one place I had to edit K1.1's files.** Two of their tests assert that
`storage` is refused *because its scenarios have not landed*. They have landed,
so the reason moved and the refusal did not. `Family::is_implemented` is now
true for all three, `FamilyRefusal::NotYet` keeps its place for the next family
that arrives as a skeleton, and a new `FamilyRefusal::OtherRunner` is what
`storage` produces when an engine runner is asked for it: still a hard refusal,
still non-zero, and the message now names `--bin storage` instead of naming this
lane. The two tests were retargeted to that, with a control asserting the other
two families still resolve so the refusal is about this family and not about
`resolve` refusing everything. Nothing else of K1.1's changed.

**Both suites on the merged tree**, which is the point of the exercise:

| suite | result |
|---|---|
| `storage_family` (the two new ones) | 2 passed |
| `storage_document` | 8 passed |
| `storage_isolation` | 3 passed |
| `families` (K1.1) | 11 passed |
| `encoding_claim` (K1.1) | 3 passed |
| `engines_family_end_to_end` (K1.1) | 3 passed |

`cargo test --no-fail-fast` over the whole merged tree: one target failed,
`version_matrix`, for the same reason as before and not for a reason belonging
to this wave. Three of its tests run `git worktree add` in the engine checkout,
which in my container is a `git archive` export with no `.git`, so they die on
"fatal: not a git repository". Every other target passed.

**A methodology finding worth more than the row it came from.** The flat-path
mutation came back GREEN the first time and it was not a surviving mutation, it
was a stale build. The revert of the previous row and the write of this one
landed inside one filesystem timestamp, cargo called the target fresh, and the
verdict was the unmutated binary's: the log has no `Compiling libviprs-bench`
line in it at all. Re-run with a real rebuild it is RED. A mutation row that
compiled nothing proves nothing and it is the same colour as one that passed, so
the runner now refuses to call such a row anything but `INCONCLUSIVE`, and I
audited every row already published: each RED carries a rebuild, and so does the
one real GREEN. Worth stealing for any lane running a mutation table between
reverts.

One thing I did not touch: `cargo clippy -- -D warnings` reports a
`collapsible_if` in `build.rs`. That file is byte for byte `origin/main`, nobody
in this wave changed it, and CI's clippy step carries no `-D warnings`, so it is
neither mine nor forced. Nothing this wave wrote draws a diagnostic.

Still not merged, and `closingIssuesReferences` is still exactly `[65]`.

PR #71, opened 2026-09-14
K1.1: families, with libviprs first and libvips behind its feature

### LINKED ISSUE #64: K1.1: families, with libviprs first and libvips behind its feature
The crate is framed end to end as a libvips comparison: the README opens with "comparing its three
pyramid engines against libvips dzsave", the report puts libvips in the same table as the three
libviprs engines, and every CI job installs libvips before it can check anything. That makes the
libviprs-only question (how do monolithic, streaming and mapreduce compare to each other, and how
does that move across releases) a by-product of a comparison it does not need.

Turn it around. libviprs-only is the default build and the default output. The comparison stays,
behind the `libvips` feature it already has.

## Families

Introduce a family concept with three members and make it the unit the runner, the report and the
charts are keyed by:

- `engines`: monolithic against streaming against mapreduce. No `libvips` feature, no FFI, no
  libvips in the container.
- `storage`: PMTiles against a directory tree. Lands in K1.2, but the family enum and the dispatch
  belong here so K1.2 only adds a member.
- `vips`: the existing comparison, gated on `feature = "libvips"`, keeping its pinned Dockerfile
  stage, `libvips-rs` FFI, provenance pin check and upstream pin validator exactly as they are.

`run-bench.sh` grows a family argument and defaults to a libviprs-only family. A build with no
features must produce a complete `engines` run with charts, which it cannot do today.

## CI

The `check` job's default cell drops its libvips install (that is the job the Hook Mirror in
`libviprs-tests/tools/install-hooks.sh` mirrors, so any job list change goes there in the same
wave). The libvips cells and the `test` job keep their install and stay feature-gated.

## Tests first

- `a_default_build_runs_the_engines_family_end_to_end`: build with no features, run the family,
  assert a snapshot with all three libviprs engines and charts on disk. RED today, where the report
  binary cannot assemble a run without a libvips row.
- `the_engines_family_never_links_libvips`: assert `cargo tree -e normal` with default features has
  no `libvips-rs`, and that no `engines` code path is reachable from a `cfg(feature = "libvips")`
  item. RED against a family that keeps the dep in the normal graph.
- `the_vips_family_is_refused_without_its_feature`: asking for `vips` on a default build exits
  non-zero with a message naming the feature. RED against a silent empty run.
- `every_family_names_itself_in_its_snapshot`: RED against an unlabelled snapshot, which is what
  makes two families' history mergeable by accident.
- `readme_and_the_article_agree_on_whether_tiles_are_encoded`: the README says every engine writes
  PNG to disk, the site article says neither side encodes. One is wrong, no guard exists, and this
  is the guard. RED today whichever way the code turns out to say it.

## Not in scope

The storage family's scenarios (K1.2, K1.4), provenance (K1.3) and anything on the site (K1.5).

spdrman:
Closes #64

The crate was framed end to end as a libvips comparison. The README opened with it, the
report put libvips in the same table as the three libviprs engines, the engine list was
built from whether a `vips` binary happened to be on `PATH`, and every CI job installed
libvips before it could check anything. That made the libviprs-only question a by-product
of a comparison it does not need, and it meant the same command measured three engines on
one machine and four on another.

A run now measures one family, and the family is the unit the runner, the report and the
charts key on.

| Family | Engines | Needs |
|---|---|---|
| `engines` (default) | monolithic, streaming, mapreduce | nothing: no cargo features, no FFI, no libvips in the image |
| `storage` | the same three, over a varying sink | nothing; the cells land in K1.2, and asking for it today is refused |
| `vips` | the three plus libvips `dzsave` | `--features libvips`, and the pinned libvips the Docker stage builds |

The engine set is a pure function of the family. It is not conditioned on
`cfg(feature = "libvips")` and not on `vips_available()`, so an `engines` run on a box
with libvips installed measures the same three engines as one on a box without, which is
what makes two runs of a family comparable at all.

### What changed

`src/family.rs` is new and holds the enum, the dispatch, the refusals and the per-family
report directory. `report`, `scalability`, `version_matrix` and `cross_version` all take
`--family`, default to `engines`, and refuse what they cannot run with a non-zero exit and
a message naming the feature or the lane that fills it in. `run-bench.sh` grows
`--family` and picks the cargo features and the Docker stage from it.

Each family writes into `report/<family>/`, its snapshot records which family it came out
of, and `push_snapshot` refuses to append one family's snapshot to another's history. Two
families in one history file is not a trend: `engines` measures three engines and `vips`
measures four on the identical sizes, so a chart drawn across both has libvips blinking
in and out between runs and a verdict dividing by a row that is only sometimes there.
Pre-family history migrates to `vips`, because the one benchmark this crate had was the
comparison.

The Dockerfile grows an `engines` target: Rust, the two crates, and deliberately nothing
else. No meson, no ninja, no multi-minute libvips compile, no codec `-dev` set. The
absence is the point, and it is also a much cheaper image. The comparison stage is
unchanged apart from its Rust pin (see the fold-ins below).

CI's `check` job now runs its default cell **before** installing libvips, and the `test`
job gains a default cell that also runs before the install. The ordering is the check:
"the engines family needs no libvips" and "the runner has libvips and we do not happen to
call it" are different claims, and only the first one is worth having. The default cell is
also where the `cfg(not(feature = "libvips"))` tests live, which until now compiled in the
check job and ran nowhere.

### The encoding contradiction

The README said every engine writes PNG tiles to a real on-disk sink; the site article
says neither side encodes and libviprs writes to a `MemorySink`. The code settles it and
the README was right. There is now one source for the sentence (`TILE_ENCODING_CLAIM`),
a test that proves it from a real run by reading the PNG signature off every tile file
each engine writes, and a guard holding every document in this repository to it. The site
half is K1.5's repository; the exact replacement paragraphs are in a comment below.

### Mutation table

Every row below was run. Each mutation was applied to the worktree, the one test it is
meant to redden was run in a container, and the file was restored from a byte snapshot taken
before the edit (never `git checkout --`, which would restore to HEAD and eat the lane's
uncommitted work). The verdicts come out of `mutations.tsv`, which the runner writes from the
actual exit code, and I checked every log for the failing test's name and message rather than
trusting the exit code alone.

One row is not what it looks like and says so: making `libvips-rs` non-optional is refused by
cargo itself (`dep:libvips-rs` requires an optional dependency), so the manifest never parses and
the test never runs. The mutation is caught, but not by this test. The row under it is the same
property mutated in a way cargo accepts, and that one is a real assertion failure.

| Test | Mutation | Observed |
|---|---|---|
| `families::the_engines_family_measures_exactly_the_three_libviprs_engines` | `Family::engines()` pushes `Engine::Libvips` when `vips_available()`, the rule the old `report` binary used, run in a container that HAS the `vips` binary | **RED(exit=101)** |
| `families::the_engines_family_is_the_same_set_with_and_without_the_libvips_feature` | the engines arm gains a `#[cfg(feature = "libvips")]` push, run with `--features libvips` | **RED(exit=101)** |
| `families::the_engines_family_never_links_libvips` | `libvips-rs` loses `optional = true` in Cargo.toml | **RED(exit=101)** (cargo-refused) |
| `families::the_engines_family_never_links_libvips` | the `default` feature list turns `libvips` on, which is the realistic way `libvips-rs` reaches the normal graph (the mutation above is refused by cargo before any test runs) | **RED(exit=101)** |
| `families::the_vips_family_is_refused_without_its_feature` | `Family::is_compiled_in()` answers `true` for `Vips` regardless of the feature | **RED(exit=101)** |
| `families::an_unknown_family_is_refused_naming_the_known_ones` | `Family::resolve` falls back to the default family on an unknown name instead of refusing | **RED(exit=101)** |
| `families::the_storage_family_exists_and_refuses_until_its_scenarios_land` | `Family::is_implemented()` answers `true` for `Storage` | **RED(exit=101)** |
| `families::the_default_family_is_libviprs_only_and_needs_no_features` | `DEFAULT_FAMILY` becomes `Family::Vips` | **RED(exit=101)** |
| `families::every_family_names_itself_in_its_snapshot` | `create_snapshot_for` writes an empty `family` field | **RED(exit=101)** |
| `families::a_history_refuses_a_snapshot_from_another_family` | `push_snapshot` stops comparing families and just appends | **RED(exit=101)** |
| `families::pre_family_history_migrates_to_the_libvips_comparison` | `migrate_snapshot` stops stamping pre-schema-3 snapshots with the legacy family | **RED(exit=101)** |
| `families::each_family_writes_into_its_own_report_directory` | `Family::report_dir` returns the report root unchanged, so all three families share one directory | **RED(exit=101)** |
| `engines_family_end_to_end::a_default_build_runs_the_engines_family_end_to_end` | the `report` binary ignores `--report-dir` and always writes to its own `report/<family>/` | **RED(exit=101)** |
| `engines_family_end_to_end::the_vips_family_is_refused_by_the_report_binary_without_its_feature` | the refusal prints its message and exits 0 instead of the refusal's own code | **RED(exit=101)** |
| `encoding_claim::every_engine_writes_encoded_png_tiles_to_a_real_on_disk_sink` | `engine_fs_sink` writes `TileFormat::Raw`, which is the site article's story: tiles on disk, no encoding | **RED(exit=101)** |
| `encoding_claim::readme_and_the_article_agree_on_whether_tiles_are_encoded` | the article's sentence "Neither side encodes to PNG or JPEG." is pasted into the README | **RED(exit=101)** |
| `encoding_claim::readme_and_the_article_agree_on_whether_tiles_are_encoded` | the Dockerfile header loses the encoding claim | **RED(exit=101)** |
| `encoding_claim::the_tile_encoding_claim_has_exactly_one_source` | the claim is typed out a second time in `src/lib.rs` rather than quoted from the constant | **RED(exit=101)** |

### Folded in rather than filed

Five things I found on the way and folded in here rather than filing:

1. **The pinned Rust could not build the thing it measures.** The Dockerfile pinned
   `rust:1.89-bookworm` and the manifest declared `rust-version = "1.85"`, while libviprs
   0.4.0 declares 1.97. Cargo refuses the build outright below a path dependency's floor,
   so `run-bench.sh`'s Docker path could not compile on `main` at all — the first thing I
   hit in a container. Both moved to 1.97, the base image re-pinned by digest. Bumping a
   measurement pin does invalidate cross-run comparison; that is the cost of the core
   moving its floor, not a choice this lane made.
2. **The libvips equivalence reference was built whatever the engine set.**
   `run_isolated_suite` built a `vips dzsave` reference pyramid per size before it knew
   whether any row would be scored against it, so a libviprs-only run on a box with `vips`
   installed still shelled out to it. Now gated on the suite actually measuring libvips.
3. **`cross_version` read a history file that is nobody's after the split.** It takes
   `--family` like everything else and reads that family's history.
4. **The `cfg(not(feature = "libvips"))` tests ran nowhere.** CI's `test` job only ran the
   libvips cell, so anything gated to the default build compiled in `check` and was never
   executed. That is where the "refuse the vips family" tests have to run, so the `test`
   job now has a default cell.
5. **Two prose guards over one file collided.** My first push put the phrases that
   contradict the encoding claim next to the claim in `src/lib.rs`, and
   `tests/vips_ffi.rs` scans that exact file for an in-RAM sink type by name as its own
   fairness guard — so quoting the wrong story there failed a different guard with a
   quotation. The negative list moved into the test that uses it. Worth knowing before
   anyone writes the third prose guard over `src/lib.rs`.

### Not measured

Nothing in this lane measures anything. Every run above is a build or a test, and the
end-to-end run is one 512x384 image at one iteration, which exists to check the shape of
the output and not its numbers.


spdrman:
### The README/article contradiction, and the half of it that is not mine to fix

The README said every engine writes PNG tiles to a real on-disk sink. The benchmark
article on libviprs.org says the opposite, twice:

> libvips writes raw tiles to a tmpdir (the minimum `dzsave` allows); libviprs writes
> to a `MemorySink` (in-memory collection). Neither side encodes to PNG or JPEG — this
> is pure pyramid generation throughput.

> `vips_dzsave` writes raw tiles (no encoding) to a temporary directory, while libviprs
> engines write to a `MemorySink`.

The code settles it, and the README was right. `BENCH_TILE_FORMAT` is `TileFormat::Png`,
every timed libviprs cell runs through an `FsSink` rooted in a real temp directory, and
both libvips paths are handed the matching `--suffix .png`. The article is prose from
before issue #153, which is the change that put both sides on the same codec, and nobody
went back to it.

`tests/encoding_claim.rs` now proves that from a real run (it writes a pyramid with each
of the three engines and reads the PNG signature off every tile file), and then holds
every document in this repository to the same sentence, refusing the phrases the stale
story is written in. What it cannot reach is libviprs-org, which is K1.5's repository, so
the site half is still open. The sentence the article should carry is
`TILE_ENCODING_CLAIM` in `src/lib.rs`, and the two paragraphs read:

> ... we call both libraries directly through `libvips-rs` FFI bindings —
> `vips_image_new_from_memory` plus `vips_dzsave` for libvips, the native engines for
> libviprs — starting from an identical `RGB8` pixel buffer (a synthetic gradient)
> already resident in memory. From there every engine writes its tiles as PNG files to a
> real on-disk sink under the same DeepZoom layout, so neither side gets an in-RAM-sink
> or tile-codec advantage: `dzsave` is invoked with `--suffix .png`, and the libviprs
> engines write through an `FsSink` at the same codec.

> ... `vips_dzsave` writes PNG tiles to a temporary directory, and the libviprs engines
> write PNG tiles to one of their own through `FsSink`, so the encode cost is paid on
> both sides.

spdrman:
### The Hook Mirror in libviprs-tests, and why I believe it needs nothing

`libviprs-bench`'s `check` job is the one `libviprs-tests/tools/install-hooks.sh` mirrors, so
anything I did to it has to be reconciled there in the same wave. What I did to it is a
**reorder only**: the `check (default)` cell moved above `Install libvips`, and no step's `name:`
or `run:` text changed. Nothing was added and nothing was removed.

Reading the guard (`libviprs-tests/tests/install_hooks_mirror_ci.rs` at `origin/main`):
`mirrored_commands` collects every mirrored job's commands into a `BTreeSet`, `hook_runs` does the
same for what the generated hook actually invokes, and `check_repo` compares them with
`difference` in both directions. Order is never observed. The `Install libvips` exemption still
matches a step the `check` job runs, so it does not go stale either.

The `test` job did grow a step (`Test (default, no libvips installed)`, with a multi-line `run:`),
but `test` is in `deferred_jobs` for this repo, not `mirror_jobs`, so `mirrored_commands` never
walks it and the multi-line rule does not reach it. No job was added or removed in either job.

**I reasoned this from the guard's source; I did not execute it.** Running it needs all seven org
repos laid down side by side with this branch in the `libviprs-bench` slot, and the local
`libviprs-tests` checkout is 56 commits behind with a dirty tree. If you want it proven rather
than argued, that is a `libviprs-tests` run against a composed tree, not something this lane can
do from its own worktree.

Issue #76, opened 2026-09-15
K2.3: vendor the causl dashboard, import and render stack as a frozen contract with a config


spdrman:
The libviprs benchmark page uses causl's dashboard, import and render stack, vendored as a frozen
copy at a pinned revision with a sync gate, which is the pattern libviprs-org already uses for the
CLI docs. The alternative, a port that drifts, is already in this repo once:
`tools/charts/chart.mjs` is a port of causl's `chart.ts` that diverged, and repeating that is
explicitly not the plan.

## Where upstream is

`/Users/rom/workspace/causl/causl-org/pages/benchmarks/` at revision `cd65c76`, readable from the
local clone. It holds `dashboard.js`, `dashboard.css`, `import-run.mjs`, `render-latest.mjs`,
`comparable-era.test.mjs`, `import-run.test.mjs`, `regen-history-sample.mjs`, `history.json` and
`history.sample.json`. **Do not reach the network for it and do not write a token anywhere.**

## What to build

`tools/contract/` holding the frozen copies, `tools/contract/UPSTREAM_REV` recording `cd65c76`, and
`tools/contract/sync-contract.sh --check` diffing the copies against that path at that revision so
drift is a red test rather than a surprise.

Then `tools/contract/config.json` carrying every libviprs-specific difference. Nothing
libviprs-specific may be edited into a frozen file. The config covers the series and their labels
and colours, the default filters, the era axes, the archive path and the family names.

## What upstream hard-codes that config has to replace

Read `dashboard.js` and find the constants: `LIBRARY_ORDER`, `LIBRARY_LABEL`, `LIBRARY_COLOR`,
`LIBRARY_DASH`, `DEFAULT_LIBRARIES`, `DEFAULT_SCALES`, `NOISY_PROXY` and the verdict thresholds.
Each needs a config path with the current literal as the fallback, so causl's own page renders
byte-identically with no config present. A golden render before and after is the test for that.

Two libviprs-specific behaviours the config must express, both of which upstream lacks:
`gated: false` on a sample renders "measured, not gated" and never a verdict chip, and a delta below
the replicate spread is `noise` rather than a pass or a regression. The storage capture measured a
median replicate spread of 3.46% with one metric at 53.3%, so this is not hypothetical.

## Era axes

Upstream breaks a comparable era on the library set. libviprs needs host fingerprint and filesystem
type as well, so a run on a different machine starts a new era instead of drawing a line across two.

## Acceptance

[...]

PR #78, opened 2026-09-15
K2.4: import an archived document into the history the page reads

### LINKED ISSUE #77: K2.4: import an archived document into the history the page reads, refusing what it cannot vouch for
An archived document is the unit of truth; `history.json` is what the page reads. Something has to
turn one into the other, and it has to refuse a run rather than average it in.

## What to build

An importer that takes an archived document, checks it may be published, and appends one compact
entry per line to `history.json`, keyed by run id. Two families now feed it, `storage` and
`engines`, and the page shows both, so the entry has to say which family it came from and the page
has to keep them apart.

Take causl's `import-run.mjs` as the base through the frozen contract copy from K2.3 rather than
writing a new one, and express every libviprs difference as config.

## What it must refuse, and why each one is real

- **An emulated or unknown run.** The probe's verdict is in the document. The published PMTiles
  numbers this epic replaced were almost certainly Rosetta and nothing recorded it.
- **A document whose digests do not verify**, and a document that is not archived at all. A number
  nobody can tie to a file is not evidence.
- **A non-publishable profile.** `ci` proves the harness runs and is never a measurement. It is
  currently admissible and archives indistinguishably from `full`, which means a smoke run can sit
  in the same era as a calibrated sweep.
- **An invariant that moved within one commit.** That is a defect, not a delta. `filesystem_entries`,
  `tiles_produced` and `output_bytes` reproduced byte for byte across every export this epic has
  seen, so a change in one at a fixed commit means something is wrong rather than something is
  slower.
- **A debug build.** Already refused at the archive door; do not let it back in here.

## What it must carry through rather than drop

The replicate spread, because the page draws bands from it and calls a delta `noise` when it is
covered. The `confidence` and `lowConfidenceReasons` on every cell, because 106 of 332 cells in the
first real capture were low-confidence for timer saturation and the page must not present those as
firm. The `gated` flag, so an ungateable metric never gets a verdict chip. And the invariants, which
are the epic's actual claim and are exact.

## The one number that matters most

The first capture shows one archive entry against 22127 filesystem entries for the same pyramid,
with the two backends' `output_bytes` within 0.016%. That is an equality table, not a chart, and the
importer has to keep it that way.

## Acceptance

- Importing the archived run `20260914T145707Z-809ee8014d002518ce55edaceba698ca7a8b8a79-0bc00939`
  produces one entry and is idempotent.
- Each refusal above has a test that fires on a real document with that one thing changed, not on a
  hand-built fixture whose shape differs from the producer's.
- `history.json` is one entry per line, so a diff over it is readable.

That last point about fixtures is not a style note. The first admission suite in this epic was green
against a document shape the producer never emits, which is how a digest that could never move
survived review.

spdrman:
Closes #77

An archived document is the unit of truth and `history.json` is what the page reads. This lane
builds the thing that turns one into the other, and refuses a run rather than averaging it in.

It is causl's `pages/benchmarks/import-run.mjs` at `cd65c76` with every libviprs difference in
config, in the shape K2.3 published on #76. Causl's checks that have an analogue here all survive:
archived by digest, a clean tree, at least one measured cell, the contamination rule read off
`machineLoad.quiet` rather than off a consequence, every reason reported rather than the first, one
compact entry per line, and idempotent on the run id.

## What it refuses, and what each one is really about

| refusal | why it is not theoretical |
|---|---|
| `provenance.emulated` is anything but an observed `false` | the PMTiles numbers this epic replaced were almost certainly Rosetta and nothing recorded it; absent, `null` and `"unknown"` are all refused, because an unobserved run is not a native one |
| the four digests do not verify | causl matches the integrity block against the index, which catches a swapped file and not an edited one, so this side recomputes them from the document's own bytes |
| the run is not in the archive index under the id the document derives | a sealed document in a working tree is not citable |
| a non-publishable profile | `ci` is three reps of a cut-down cell list and archives indistinguishably from `full`; both the label and the resolved invocation have to say a publishable one |
| an invariant moved within one commit | a defect, not a delta. Across commits the same difference is imported as a finding, which is the control that keeps the rule from being "refuse every change" |
| a debug build | already refused at the archive door; this is the door it must not come back in through |

## What it carries through rather than drops

The replicate spread, with the cell it was measured on, so the page can draw a band and call a delta
`noise` without implying the floor was measured everywhere. `confidence` and `lowConfidenceReasons`
on every sample: 106 of the 332 measured cells in the capture are low-confidence for timer
saturation, 95 of them as samples and 11 as the replicate pass, and the page must not present any of
them as firm. `gated` with a reason attached, so an ungateable metric never gets a verdict chip. And
the invariants as an equality table with an `exact` flag, which is the epic's actual claim: **one
archive entry against 22127 filesystem entries for the same 21851-tile pyramid, with the two
backends' `output_bytes` 0.00807% apart.**

## Proof

```
docker run --rm --platform linux/arm64 -v $PWD:/w -w /w node:22-bookworm-slim \
  node --test --test-concurrency=1 tools/publish/canonical-json.test.mjs tools/publish/import-run.test.mjs
# 56 tests, 56 pass, 0 fail, exit 0

node tools/publish/import-run.mjs --document archive/storage/20260914T145707Z-...-0bc00939.json
# history 0 → 1 entries, exit 0
# again: history 1 → 1 entries, exit 0, and the file is byte-identical
```

The strongest single assertion in the suite is that the JavaScript canonicaliser reproduces the
producer's own `integrity` block over the real capture, byte for byte, all four digests. That is the
cross-language pin I2.2 wants; it arrived early because verifying a digest needs it.

## Which document shape this is built against

Both, on purpose, and the one place the two disagree is named in the config rather than chosen.

`storageAttested` becomes `attested` on `lane/k2.2-engines-document`, because a second family
arriving is what turns a field name into a family name. The archive keeps documents written before
that rename forever, so `producer.attestedFrom` lists both, newest first, and a cell carrying
**neither** is its own refusal that says the document is a shape this importer has not been taught.
Reading one name only is the interesting failure: every document of the other shape comes back
`undefined`, which reads as "unattested" and refuses the whole archive while looking exactly like a
gate doing its job. `M43` and `M44` are those two mistakes, and both are red.

Everything else on that branch lands where it already was. `measurement.reps` becomes a map and
`storage` still serialises `{"generate": 7, "read": 20}`, which is carried verbatim.
`archive/<family>/` is derived the way `archive::dir_for_family` derives it rather than hard-coded.
`backend`, `scale`, `source`, `cell` and `key` stay top-level on the cell: `CellLabels` is a builder
input and is not what gets serialised.

And `runId`, which was `null` in every document ever written in both families: this importer has
never read it. The id is derived from `startedAt`, the library commit and the host bucket, then
joined to the archive index, which is what makes a re-import replace rather than append. The
idempotency test asserts the key is the real id and not null, so it cannot pass by everything
colliding, and a document that *states* an id disagreeing with the one its evidence derives is a
refusal for when the producer starts writing one.

## The mutation table

Forty-seven mutations, every row run in the container against the code as it stands, none survived.
Each row had to produce the same red set on two consecutive runs before it was recorded, which
matters: see the findings comment.

| id | mutation | red | first test that caught it |
|---|---|---|---|
| `M1` | emulated: admit unknown and absent by only refusing an explicit true | 2 | an absent emulation verdict is refused (+1 more) |
| `M2` | digests: trust the integrity block the document was handed | 1 | a document whose digests do not verify is refused |
| `M3` | digests: spell the runners block with the key no document has | 1 | the javascript digests reproduce the producers own integrity block |
| `M4` | digests: take the document digest over the whole file, integrity included | 25 | a baseline fitted on another host or filesystem gates nothing (+24 more) |
| `M5` | archive: publish a document whose archive has no index at all | 1 | a document that is not archived is refused |
| `M6` | archive: take the run id from the index row rather than deriving it | 1 | a run id the index states but the document does not derive is refused |
| `M7` | archive: accept any archived digest without comparing it | 1 | a document whose digest is not the one the index records is refused |
| `M8` | profile: admit every profile the producer can emit | 1 | the ci profile is refused |
| `M9` | profile: read the document label and not the resolved invocation | 1 | a profile the document and its invocation disagree about is refused |
| `M10` | invariants: append every run and never compare one to the last | 1 | an invariant that moved within one commit is refused |
| `M11` | invariants: refuse any change anywhere, commit or no commit | 1 | an invariant that moved across commits is imported as a finding |
| `M12` | debug build: read provenance for the host and not for the toolchain | 2 | a debug build is refused (+1 more) |
| `M13` | debug build: accept any build profile | 1 | a debug build is refused |
| `M14` | dirty: a missing flag reads as a clean tree | 1 | a dirty flag that is absent is refused, not read as clean |
| `M15` | history: append rather than replace on the run id | 2 | importing the same archived run twice leaves one entry (+1 more) |
| `M16` | history: pretty-print the array | 1 | history is one entry per line so a diff over it is readable |
| `M17` | confidence: keep the verdict and drop the reasons for it | 1 | every sample carries its confidence and the reasons for it |
| `M18` | replicate: drop the spread, leaving the page with no noise figure | 1 | the replicate spread reaches the entry and every sample it was measured for |
| `M19` | gated: grade everything against causl’s ten percent constant | 1 | nothing is gated without a fitted baseline and a p99 is never gated |
| `M20` | gated: take a baseline fitted on another host at face value | 1 | a baseline fitted on another host or filesystem gates nothing |
| `M21` | invariants: drop the equality table because it never moves | 2 | an invariant that moved within one commit is refused (+1 more) |
| `M22` | sections: leave the source image out of the scenario | 3 | every sample carries its confidence and the reasons for it (+2 more) |
| `M23` | replicate pass: publish both measurements of the control cell | 3 | every sample carries its confidence and the reasons for it (+2 more) |
| `M24` | units: write every median into a field that says milliseconds | 1 | a sample keeps its unit and direction and never converts a rate into milliseconds |
| `M25` | outcomes: inherit causl’s rule that any failed cell refuses the run | 22 | a baseline fitted on another host or filesystem gates nothing (+21 more) |
| `M26` | cells: let an ok cell publish without an observed attestation | 1 | an unattested ok cell is refused |
| `M27` | cells: publish a sweep that measured nothing | 1 | a run with no ok cells is refused |
| `M28` | load: read confidence and never the machine load underneath it | 1 | a run measured mostly under contention is refused |
| `M29` | refusals: stop at the first one found | 4 | an absent emulation verdict is refused (+3 more) |
| `M30` | flags: treat an unknown flag as a no-op | 1 | an unknown flag is a usage error and not a silent no op |
| `M31` | family: read every document with one family’s assumptions | 1 | a document from another family or schema version is refused |
| `M32` | declared: publish a declared request count as a measured one | 1 | the declared request counts are marked declared and are never gated |
| `M33` | canonical: drop a null-valued key instead of writing it | 5 | absent and null digest differently (+4 more) |
| `M34` | era: report the archive bucket as the host fingerprint | 1 | the entry carries the era axes the page breaks a line on |
| `M35` | dry run: write the history anyway | 1 | a dry run writes nothing |
| `M36` | outcomes: treat an outcome nobody classified as a silent skip | 1 | an outcome the config does not classify is refused, not silently skipped |
| `M37` | invariants: carry a name nobody classified without saying so | 1 | an invariant name the config does not classify is refused |
| `M38` | skips: give an impossibility and a self-refusal the same label | 1 | a non ok cell becomes a skip with its reason and does not refuse the run |
| `M39` | runner: let the runner and the family disagree | 1 | a runner that disagrees with its family is refused |
| `M40` | fingerprint: concatenate the fields instead of separating them | 23 | a baseline fitted on another host or filesystem gates nothing (+22 more) |
| `M41` | cell shape: read a document of another shape with this one’s assumptions | 1 | a measured cell missing a field the config names is refused |
| `M42` | refusals: print the reasons and carry on importing anyway | 25 | a cell attested under the new field name imports, and under neither is refused (+24 more) |
| `M43` | attestation: read only the new field name, refusing every archived document | 22 | a baseline fitted on another host or filesystem gates nothing (+21 more) |
| `M44` | attestation: read only the old field name, refusing every document K2.2 writes | 1 | a cell attested under the new field name imports, and under neither is refused |
| `M45` | run id: key the entry on the one the document states | 3 | an invariant that moved across commits is imported as a finding (+2 more) |
| `M46` | run id: accept a stated id that disagrees with the derived one | 1 | the run id the document states must be the one its evidence derives |
| `M47` | cells: read the document header for dirt and never the cells | 1 | a cell stamped dirty is refused |

## Where the files are

`tools/publish/` holds the importer, the canonicaliser, the tests and a provisional config;
`archive/storage/` gains the capture itself and its index row, because a refusal test on a
hand-built fixture proves the fixture wrong rather than the importer right. The tests run in CI
through a new workflow **file**, `publish-tests.yml`, so libviprs-tests' Hook Mirror is not tripped,
which is the pattern `storage-measured.yml` already set.

Findings, disagreements and the compose steps are in a comment below rather than in new issues.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01NaKQR8REqRrFtjeS43AUTA


spdrman:
Findings from this lane, folded in here rather than opened as issues.

## One deliberate disagreement with the config schema, and it needs a decision

`producer.gatedFrom` is `null`, and the schema comment on #76 says the importer should then write
`gated: null`. The argument is sound: "nobody said" and "not gated" are different facts, and if
absent counted as ungated the page goes silent the day the field is wired up.

But `verdict.gate` only suppresses a chip on the literal `false`, so `null` hands all 284 samples to
`verdict.regressionPct`, which is a 10% constant. 136 of those samples are low-confidence, nothing
has been calibrated on this host, and the free replicate pair already moved
`directory.read_concurrent@4.lookups_per_s` by 53.3% on an idle machine. Every one of them would get
a verdict chip computed against a guess, which is the defect this epic exists to remove and the one
the plan names outright: do not put a constant threshold anywhere a verdict is computed.

So the importer writes `false` with a reason while no calibration exists, and `true` only where a
fitted baseline supplies a tolerance for that exact cell on this host and filesystem. It is a
decision rather than an absence, and it is behind `producer.gatedWhenUncalibrated`: set that to
`null` and you get the schema comment's literal behaviour. Worth settling between K2.3, K2.5 and the
calibration lane before the page ships, because it decides whether the first published page carries
284 chips nobody fitted.

## Two tests were passing for the wrong reason, and the mutation table is how I know

Deleting the archive-index check outright left the suite green, and so did deleting the
publishable-profile check. Both tests were asserting `exit 1` plus a regex, and with the check gone
the importer **threw**: an uncaught exception is also exit 1, also on stderr, and its stack trace
happened to contain the word each test was matching on. One was matching `/index\.json/` against a
stack frame naming the file it failed to open, the other was being refused by the
profile-agreement check instead of the one it was about.

Every refusal test now goes through a helper that asserts the refusal banner and the *absence* of a
stack frame, which a crash cannot satisfy in either direction. The `ci` test also sets
`provenance.invocation.resolved.profile`, so only the check it is about can refuse it.

## A mutation table run forty containers deep produces spurious kills

Under back-to-back container load `spawnSync` returns `status: null` with an `error`. The test read
that as "the importer did not exit 0", which is indistinguishable from a refusal, and one mutation
was credited with killing twelve tests when it kills one. Two separate shapes: the whole test file
failing to run (`# tests 2`), and individual spawns failing inside a run that otherwise worked.

Fixed on both sides. The harness retries a failed spawn once and then fails saying the *harness*
failed. The driver pins the expected test count and requires two consecutive runs to agree on the
same red set before it records a row; it caught five disagreements in the final pass, including a
"31 vs 21" that would have gone into the table as fact. Every number in the table above is from a
run that agreed with its neighbour.

## Three things the real document forced that a fixture would not have

**causl's "no cell may be failed" rule refuses this capture.** Seven cells are `failed` because a
directory tree has no root directory to decode, which is a fact about the backend rather than a
broken run, and four more are `refused` because `allocated_bytes` moved between reps of one
generation. Under upstream's gate the whole document is unpublishable. K2.3's three outcome buckets
handle it; what survives from causl is the run-level form, that at least one cell must have been
measured. The two kinds carry a `kind` field so the page can tell an impossibility from a
self-refusal, and an outcome the config does not classify is a refusal rather than a silent skip.

**The replicate cell is in the document twice.** Every `full` sweep measures the 93 cell first and
last, so 48 of the 332 measured cells are a second pass. Publishing both draws one run as two points
and makes the control itself look like a regression, so the second pass goes into `replicates[]`
beside the samples and the spread it was taken for is already in `replicate.spreadPct`.

**Dropping `source` from the section key collides 94 real sections.** The test proves that rather
than asserting it: it builds both keys and counts. 21851 is `8192x8192@64+gradient` and
`8192x8192@64+noise`; 16369 is two `4096x6256@46` cells.

## Smaller things

- **Two fingerprints, not one.** `producer.hostFingerprintFrom` (the page's era axis) and the
  producer's own `host8` (the last segment of every run id) are different lists over the same
  provenance: `rustc` and the scratch filesystem are in one and `ncpu` is in the other. The entry
  carries both, as `host.fingerprint` and `host.archiveBucket`, because letting either stand in for
  the other silently changes what counts as the same machine. `archiveBucketFrom` is in the config
  so it stays pinned to the producer.
- **`allocated_bytes` is not exact within a commit.** It is the filesystem's answer to the same
  bytes and moves with the block size, so it is compared only between runs on the same `fsType` and
  mount source. In the exact list it would refuse an honest run on another disk.
- **The within-commit invariant check keys on both trees.** The harness generates the source image,
  so a harness-only change can move `output_bytes` honestly. The cost is that a harness-only change
  does not get the within-commit check, and the gate's own comparison is where that is caught.
- **A number above 2^53-1 is refused even where the producer would allow it.** The Rust side refuses
  a `u64` past that and allows an `f64` of 1e20; after `JSON.parse` the two are the same value and
  JavaScript cannot tell them apart, so this side refuses both. No benchmark document carries a
  number up there, and refusing cannot admit a digest the producer would have refused.
- **`medianMs` is null on a rate.** Ten of the thirty metric keys are `higher-is-better` and twelve
  are not durations at all. The document's own `median` and `unit` travel under their own names,
  `medianMs` is filled only for a duration and `throughput` only for a rate, and the section carries
  `direction`. That last one matters for the gap K2.3 flagged: the frozen dashboard has no notion of
  direction, so a `lookups_per_s` cell that got faster renders as `regressed`. The importer carries
  the fact; nothing at this pin reads it.

## Compose steps this leaves behind

1. `tools/publish/config.json` folds into K2.3's `tools/contract/config.json` and is deleted.
   `--config` already prefers the contract copy when it exists, so this is a deletion rather than an
   edit.
2. `tools/publish/history.json` moves to `libviprs-org/perfbench/history.json` at K2.5, and
   `producer.historyPath` follows it.
3. `publish-tests.yml` and K2.3's `node --test` step collapse into one, and the libviprs-tests hook
   mirror changes once instead of twice.

spdrman:
Second findings pass, after reading `lane/k2.2-engines-document` against every field this importer
reads. Folded in here rather than opened as issues.

## What I checked, and what I found

Four names moved on that branch. I went through the serialised structs rather than the archived
capture, because the capture predates them.

| name | what happened | what this importer does |
|---|---|---|
| `storageAttested` → `attested` | the rename is real, `DocumentCell.attested` | reads both, newest first, from `producer.attestedFrom`; a cell carrying neither is its own refusal |
| `measurement.reps` scalar struct → map | `Reps(BTreeMap)` is `#[serde(transparent)]`, so `storage` still writes `{"generate":7,"read":20}` | carried verbatim, unaffected |
| flat archive → `archive/<family>/` | `archive::dir_for_family` strips the `libviprs-` prefix | derived the same way, `archiveDirByFamily` overrides one family if ever needed |
| `runId` null → populated | `stamp_run_id` calls the same `archive::run_id` | never read; derived and joined to the index, and a stated id that disagrees is a refusal |

One that looked like a rename and is not: `CellLabels` gathers `backend`, `scale`, `source` and
`cell`, but it is a builder input with no `Serialize`, and `DocumentCell` still writes all four at
the top level. Nothing to change.

## The rename is the dangerous kind and it is dangerous in both directions

Reading only `attested` refuses every document in the archive today; reading only `storageAttested`
refuses every document K2.2 writes. Both are 100% refusal, both look exactly like an admission gate
doing its job, and neither says anything about the document being a shape nobody taught the reader.
That is why a cell carrying *neither* name gets its own refusal with its own wording. `M43` reds 22
tests and `M44` reds one; both are in the table.

The same reasoning produced a wider guard while I was there: every field the config names as part of
the series, the section or its axis (`backend`, `key`, `source`, `scale`, `unit`, `direction`) is now
required on a measured cell. Without it an `engines` document of another shape would have had its
section names quietly shortened by `scenarioOf` and its `direction` come out null, which is a page
that is wrong rather than a page that is missing.

## The idempotency test was not vacuous, and now proves it

`doc.runId` has been null in every archived document. This importer never read it, so the key was
always the derived id, but the test asserted only that two imports leave one entry, which is exactly
what two null keys would also produce. It now asserts the document's own field is null, the entry's
`runId` is the real id, and the second import rewrites the file byte for byte.

## Two paths that had never run before this lane

`--allow-dirty` stamps `dirty: true` on every cell and the producer has never written the field, so
the cell-level dirty refusal here was inherited from causl and unreachable. It now has a test. The
document-level `provenance.dirty` and `provenance.library.dirty` checks are the live ones, and a
missing flag is refused rather than read as clean (`M14`).

The other is the refusal gate itself. Writing the cell-shape test showed it sat *after* the entry was
built, so a document already refused for being malformed was then read by code that did not survive
it, and the crash printed a stack where the list of reasons should be. The reasons are the thing
that tells an operator how many re-runs this is going to take, so the gate moved ahead of the reader
and the invariant comparison moved up to join it (`M42`, 25 red).

## Per-engine peak RSS

Noted from K2.1: `scalability.rs` measures every cell in its own child now, so `peak_rss_mb` is real
per engine. This importer carries it through from `invariants.peakRssMb` into `samples[].peakRssMb`,
with `peakHeapMb` from `heapPeakBytes` where the counting allocator ran and `null` where it did not,
so the field names stay true. Nothing to change; the numbers are just no longer one figure repeated.

spdrman:
On the `producer.gatedFrom` question you escalated: your call is right, keep `gated: false` with the reason.

The schema comment says write `gated: null`, but `verdict.gate` only suppresses a chip on a literal `false`, so null is not neutral here. It hands all 284 samples to a 10% constant at a moment when 136 are low-confidence and the replicate pair moved one metric by 53.3%. That would put regression chips on noise, on the first page anyone sees, which is the precise failure this epic was started to stop.

The deeper reason is that **nothing is gateable yet**. A tolerance has to be fitted from repeat sweeps before a verdict means anything, and that calibration is not in this wave: the original design had it and the scope landed on the frozen contract instead. Until it exists, every cell is honestly "measured, not gated", and `false` says so where `null` quietly says "judge it against a guess".

So `producer.gatedWhenUncalibrated` stays, and it should carry its reason into the page rather than being a silent default. When calibration lands it flips to reading a fitted tolerance per cell, and the flag becomes the thing that says which cells earned one.

Two other things in your report worth recording here rather than leaving in a thread.

**Your refusal tests were passing for the wrong reason and you found it.** With the targeted check deleted the importer threw, and an uncaught exception is also exit 1 carrying the matched word in its stack trace. So the test could not tell a refusal from a crash. Asserting the banner and the absence of a stack is the right fix, and it is the same shape as every other finding this epic has turned up: a green that was green for a reason nobody had checked.

**Reading both `storageAttested` and `attested`, with a cell carrying neither getting its own refusal, is better than what I asked for.** I only told you the name had moved. Reading one name alone would refuse 100% of the other shape while looking exactly like a gate doing its job, and a rename that silently rejects everything is worse than one that silently accepts everything, because it looks like rigour.

And the `spawnSync` finding belongs with the others: under back-to-back container load it returns `status: null`, your runner read that as a refusal, and one mutation got credited with killing twelve tests when it kills one. A run that did not happen wearing the colour of one that did, for the fifth time in this epic and the second time inside a mutation runner.

PR #79, opened 2026-09-15
K2.3: freeze causl's dashboard stack at cd65c76 and drive the differences from config

### LINKED ISSUE #76: K2.3: vendor the causl dashboard, import and render stack as a frozen contract with a config

spdrman:
Closes #76

causl's dashboard, importer and renderer come into `tools/contract/` as a frozen copy at `cd65c76`
with a sync gate, and every libviprs difference is a value in `config.json` rather than an edit to a
frozen file.

The reason for the pattern rather than a port is already in this repo once. `tools/charts/chart.mjs`
started as causl's `chart.ts` and is no longer one, so the two cannot be reconciled and a fix to
either does not reach the other.

## What is here

`tools/contract/upstream/` holds the nine files byte for byte, and `UPSTREAM.manifest` records the
blob sha1 `git ls-tree` reported for each at the pin.

`sync-contract.sh --check` runs two modes and at least one of them always runs. Manifest mode
re-hashes every frozen file the way git hashes a blob and needs neither git nor a clone, so it works
in CI and in the node container, where the causl repositories (private Gitea) are unreachable. Clone
mode re-derives each sha1 from a local causl-org clone, so a manifest doctored to match a doctored
frozen file still fails, which is mutant M3 below. A missing clone downgrades to manifest-only and
says so; an unreadable manifest refuses with exit 3 rather than reporting green.

`parameterize.mjs` rewrites the frozen dashboard into a config-driven one. Each of its 18 edits is
anchored on an exact string and must match exactly once, so upstream moving a line is a refusal, not
a patch landing somewhere else. Every hard-coded constant becomes `__cfg('<path>', <the same
literal>)`, and the two behaviours libviprs needs are config-gated and default off.

## The golden render

With no config present the parameterised dashboard emits exactly what the frozen one does, and that
is the property that would make this parameterisation safe to offer upstream: it changes nothing
about causl's own page.

Both files render through a 150-line DOM shim over causl's sample history, causl's real history
(which spans two library sets, so the era windowing path actually runs) and the 404 fallback path,
and the HTML has to match byte for byte. There is a positive control: loading `config.json` must
CHANGE the render, or the equality tests are comparing two things that cannot differ.

The shim is not jsdom because the comparison is between two renders of the same page, so anywhere
the shim is less than a browser both sides are equally less than a browser, and the test gains no
dependency, no lockfile and no install step. What it cannot show is that the page looks right in a
browser, and it does not claim to.

The render is stable on node 20, 22 and 26, checked, so the golden is not pinned to one runtime.

## The two behaviours upstream lacks

`gated: false` on a sample renders "measured, not gated" and never a verdict chip, because a chip is
a claim that a threshold ruled on the number and when nothing gated the cell no threshold did. Only
the literal `false` counts: absent and null mean nobody said, which is a different fact.

A delta smaller than the cell's own replicate spread is `noise`, checked before the regression and
improvement lines. Run `20260914T145707Z` measured a median replicate spread of 3.46% across 48
metrics with `directory.read_concurrent@4.lookups_per_s` at 53.3%, so on that metric upstream's 10%
line rules a 20% move a regression and a 20% move is what an idle machine hands you. A cell carrying
no measured spread is `unknown` rather than falling through to the thresholds: a spread that was not
measured is not a spread.

Both are rendered against `fixtures/history.libviprs.json`, whose first run is the archived capture
cell for cell and whose second run is constructed, moved by hand to reach each verdict kind, and
says so in every entry.

## Era axes

Upstream breaks an era on the measured library set alone, which is right when every run comes off one
CI machine. libviprs adds the host fingerprint and the filesystem type, so a run on a different
machine starts a new era instead of drawing a line across two. With the default single axis the
signature agrees with upstream's `libsOf`/`sameLibs` on every entry of the shipped history, which is
asserted rather than assumed.

## Findings folded in rather than filed

Six things turned up that were not in the issue. They are in a comment on this PR: the `failed`-cell
refusal that would reject the whole real capture, the non-unique tile count, the chart-point trap
that made the noise rule silently inert, the direction gap, the 106-of-332 timer-saturated split, and
the repository size this adds.

## Gate

57 tests: 34 contract tests plus causl's own 23 running against the vendored copy. 21 mutants, 0
escaped. Everything ran in `node:22-bookworm-slim` on `--platform linux/arm64`; nothing touched the
host toolchain.

```
docker run --rm --platform linux/arm64 -v "$PWD":/repo -w /repo/tools/contract \
  node:22-bookworm-slim npm run gate
```


spdrman:
## Findings folded into this PR rather than filed as issues

Six things turned up while checking the config against the archived run
`20260914T145707Z-809ee8014d002518ce55edaceba698ca7a8b8a79-0bc00939`. None of them is speculative;
each is asserted by a test in this PR so it cannot quietly stop being true.

**1. Upstream's importer would refuse the entire real capture.** `import-run.mjs` gate 4 is "no cell
may be `failed`", and seven cells in the capture are `failed` with the reason `a directory tree has
no root directory to decode; its index work happens in the kernel one path resolution at a time`.
That is a fact about the backend, not a broken run. Upstream's own frozen test says so out loud: `a
failed cell is refused and named`, and it passes against our vendored copy. So `producer.outcomes`
has three buckets rather than two, `failed` and `refused` sit in `structural` and import as skips
carrying their reason, and `refusing` is empty. K2.4 has to honour that or the first import refuses.
Pinned by `contract.test.mjs` "the real capture carries structurally-failed cells".

**2. Tile count is not unique across cells.** `21851` is both `8192x8192@64+gradient` and
`8192x8192@64+noise`; `16369` is both `4096x6256@46` cells. A section keyed on tile count alone
draws gradient and noise as one line and nothing on the page says so. `sections.scenarioFrom` is
therefore `["key","source"]` and `scaleFrom` stays `scale`, which is only sound because `source` is
in the scenario. Pinned by "tile count is NOT unique across cells".

**3. A chart-point trap that made the noise rule silently inert.** `groupSections` builds a
fixed-shape point object and that object is the only thing `computeVerdict` ever sees, so a verdict
rule reading a sample field reads `undefined` until the field is carried across. A rule that reads
`undefined` looks exactly like a rule that is switched off, and this cost an eighteenth anchored
edit (`carry-fields`) that nobody would have guessed at from the issue. `samples.carry` is the
config path; anything K2.5 wants on a point goes in it. Mutant M9 removes the edit and the noise
test goes red.

**4. The frozen dashboard has no notion of metric direction.** Ten of the thirty metric keys are
`higher-is-better`, and for those a cell that got FASTER renders as `regressed` because its number
went up. `sections.directionFrom` carries the fact and nothing at this pin reads it. I did not fix
it: fixing it means a real edit to the verdict logic and that belongs with whoever owns the page
layout (K2.5). There is a test pinning the current behaviour so the gap turns up in CI rather than
on the published page.

**5. 106 of 332 measured cells are timer-saturated, and 159 are low-confidence.** The issue quotes
106, which is exactly the timer-saturated subset: `timerTickNs` 41 times `minTicksPerSample` 100 is
the 4.1 us floor. The other 53 are low-confidence for CoV above 0.15, which is a different reason and
wants different handling on the page. `verdict.observed` records both counts and derives the 4.1 us
floor from the tick rather than carrying it as a typed number, asserted by "verdict.observed matches
the run it cites".

**6. This adds about 3.3 MB to the repository.** `upstream/history.json` is 2.9 MB of causl's own
measurements and `golden/dashboard.sample.html` is 320 KB. The history is not decoration: causl's
`comparable-era.test.mjs` asserts against it, and without it that frozen test cannot run at all.
Flagging it rather than deciding it. If it is not worth 2.9 MB, the thing to drop is
`upstream/history.json` plus that one frozen test, and the sha256 in `golden/RENDERS.sha256` for the
real-history render goes with it.

## The mutation table

Every row was run. Applied one at a time against a snapshot, the named test run, then restored.

| # | mutation | caught by | exit |
|---|---|---|---|
| M1 | `upstream/dashboard.js`: `NOISY_PROXY 0.50` → `0.42` | `sync-contract.sh --check` | 1 |
| M2 | `upstream/libviprs-tweak.js` added to the frozen directory | `sync-contract.sh --check` (UNPINNED) | 1 |
| M3 | frozen file edited AND the manifest doctored to match | `sync-contract.sh --check`, clone mode only | 1 |
| M4 | `parameterize.mjs`: the `NOISY_PROXY` anchor renamed so it matches nothing | `contract` "an anchor that moved is a refusal" | 1 |
| M5 | `parameterize.mjs`: `__verdictChip` emits a different class name | `golden-render` "byte for byte the same" | 1 |
| M6 | `parameterize.mjs`: the noise rule defaults ON instead of OFF | `golden-render` "byte for byte the same" | 1 |
| M7 | `parameterize.mjs`: `__gateVerdict` treats anything not `true` as ungated | `verdict-behaviour` "only the literal false is ungated" | 1 |
| M8 | `parameterize.mjs`: the noise check moved after the regression check | `verdict-behaviour` "a delta inside the cell replicate spread is noise" | 1 |
| M9 | `parameterize.mjs`: the `carry-fields` edit dropped | `verdict-behaviour` "a delta inside the cell replicate spread is noise" | 1 |
| M10 | `config.json`: `onMissingSpread` `unknown` → `skip` | `verdict-behaviour` "a cell with no measured spread is unknown" | 1 |
| M11 | `config.json`: the `noise` colour removed while the rule stays on | `contract` "passes the structural validator" | 1 |
| M12 | `config.json`: `series.order` loses `directory` | `contract` "names only things the producer actually emits" | 1 |
| M13 | `config.json`: `failed` moved into `producer.outcomes.refusing` | `contract` "structurally-failed cells" | 1 |
| M14 | `config.json`: `contract.upstreamRev` points at another revision | `contract` "passes the structural validator" | 1 |
| M15 | `generated/dashboard.js` hand-edited | `parameterize.mjs --check` | 1 |
| M16 | `golden/dashboard.sample.html` hand-edited | `golden-render` "the committed golden still matches" | 1 |
| M17 | `config.json`: `era.axes` drops the host and filesystem axes | `contract` "a run on a different host or filesystem starts a new era" | 1 |
| M18 | `lib/config.mjs`: `UPSTREAM_DEFAULTS verdict.passPct 0.05` → `0.06` | `contract` "UPSTREAM_DEFAULTS still matches the literals" | 1 |
| M19 | `lib/dom-shim.mjs`: the serializer stops emitting style properties | `golden-render` "the shim renders the things the parameterised constants feed" | 1 |
| M20 | `sync-contract.sh`: the unpinned-file loop deleted | `contract` "red against an unpinned file" | 1 |
| M21 | nothing mutated (the negative control) | the whole gate | 0 |

21 mutants, 0 escaped.

M3 is the one worth reading twice. Manifest mode reports `dashboard.js ok` because the manifest was
doctored to agree with the edit; clone mode reports `DIFFERS from cd65c76` because it re-derives the
number from the clone rather than trusting the manifest's copy of it. Both lines appear in the same
run, which is how you can tell the two modes are actually independent.

## Not done, and said so rather than implied

`import-run.mjs` and `render-latest.mjs` are frozen and vendored but NOT parameterised. The config
carries everything the importer needs (`producer.*`, `sections.*`) and K2.4 owns the importer; the
renderer's causl-specific prose is K2.5's call. Nothing in this PR reads either file at runtime.

`producer.gatedFrom` is `null` because no cell in the archived capture carries a `gated` field.
The ungated behaviour is proven against a constructed fixture value, not against a producer that
emits one. When a producer starts emitting it, set that path and the behaviour is already wired.

No browser was opened. The shim is not a browser and the golden proves equivalence between two
renders, not that the page looks right. That check belongs to K2.5 with the real page.

There is no real libviprs `history.json` yet, so nothing here has been rendered from an imported
run. That is K2.4's output and this contract is what it imports into.

One transient: a single run of `contract.test.mjs` reported the sync-gate test red immediately after
a host-side file write, and the same file passed on the next two runs and every run since. It looks
like Docker Desktop bind-mount propagation rather than the gate; I did not chase it further.

PR #80, opened 2026-09-15
Measure each engine's peak RSS in its own child, in the sweep that publishes

### LINKED ISSUE #74: The engines sweep reports one peak RSS for all three engines, because it reads a process-wide watermark
The first full `engines` capture reports **byte-identical peak RSS for all three engines, in all
twenty (megapixel, concurrency) groups**. At 188.7 MP every engine reports `1219.7734375` MB, to
seven decimal places. Three engines with very different memory models do not agree to the fraction
of a megabyte twenty times running.

Meanwhile `tracked_memory_mb`, the engine's own accounting, diverges exactly as it should:

| MP | monolithic | streaming | mapreduce |
|---|---|---|---|
| 47.2 | 135 MB | 21 MB | 21 MB |
| 188.7 | 540 MB | 39 MB | 39 MB |
| 280.0 | 801 MB | 49 MB | 49 MB |

So the memory story is real and only one of the two columns is telling it.

## Where it comes from

`scalability.rs` measures with `process_peak_rss()`, which is
`getrusage(RUSAGE_SELF).ru_maxrss`. That is a process-wide high-water mark and it never comes back
down. The binary runs all three engines in one process: `grep -nE 'Command::new|fork|spawn|wait4|current_exe'`
over `src/scalability.rs` finds nothing but a doc comment. So whichever engine peaks highest, which
is monolithic because it holds the whole canvas, sets the watermark, and every engine measured after
it reports that number as its own.

## This is issue #157, still open in the binary that publishes

`tests/rss_isolation.rs` opens by describing this defect in its own words:

> Peak RSS used to be `getrusage(RUSAGE_SELF).ru_maxrss`, a monotonic, process-wide high-water mark
> shared by every in-process engine. In that world a SMALL-image run performed after a LARGE-image
> run reported the LARGE run's peak, because the watermark never comes back down. Every memory
> number was therefore contaminated by whatever ran earlier. The fix runs each cell in its own child
> process and reads that child's `ru_maxrss` via `wait4`.

The fix landed in `libviprs_bench::harness::spawn_single_cell`, and the test guards
`spawn_single_cell`. `scalability.rs` never adopted it. So the regression test for this bug is green
while the bug is live in the binary the `engines` family runs and the site's charts are built from.

The field's own doc comment calls it "the cross-engine-comparable memory basis", and
`tiles_per_second_per_mb` and `resource_cost` are both derived from it, so two more published
columns inherit the contamination.

## What to change

`scalability.rs` measures each engine through `harness::spawn_single_cell`, the way the `report`
binary already does, so every engine's `ru_maxrss` is its own child's.

## Acceptance

- Peak RSS differs between monolithic and streaming at a size where their tracked working sets
  differ by an order of magnitude. Red today: they agree to seven decimal places.
- The regression test covers the path that publishes. Today it covers only the path that was fixed,
  which is why this survived.
- `tiles_per_second_per_mb` and `resource_cost` are recomputed from the per-child figure.

## What not to do

Do not reorder the engines so monolithic runs last. That makes the numbers look plausible without
making them measurements, and the next person to add an engine puts it back.

## How it was found

Running the capture, not reading the code. The numbers being identical is what showed it: no test
asserted that three engines produce three different peak RSS figures, because the test that would
have is pointed at the other code path.

spdrman:
Closes #74

The `engines` sweep published the same peak RSS for all three engines, in all
twenty (megapixel, concurrency) groups, to seven decimal places. `scalability.rs`
measured with its own `process_peak_rss()`, which is
`getrusage(RUSAGE_SELF).ru_maxrss`: a process-wide watermark that never comes
back down, read by three engines running in one process. Monolithic holds the
whole canvas, so it set the watermark and every engine measured after it
published monolithic's number as its own. `tiles_per_second_per_mb` and
`resource_cost` are both derived from that field, so three published columns
carried the same one number.

The fix for this shipped a long time ago. `harness::spawn_single_cell` has been
re-invoking the binary with a hidden `--single`, reading one cell's metrics back
as JSON and reaping the child with `wait4` since #157, and `report` has measured
through it ever since. `tests/rss_isolation.rs` has been guarding it just as
long. The sweep never adopted it, so the regression test for this bug sat green
while the bug ran in the binary the site's charts are drawn from...

## The change

`scalability.rs` now wires `maybe_run_single_subcommand` ahead of its own
argument parser and measures every cell through `harness::spawn_single_cell`,
exactly the way `report` does. Each `(engine, size, concurrency)` cell gets its
own process and its own `ru_maxrss`. The two derived columns follow for free,
because they were always computed from `peak_rss_mb`; what changed is what that
field holds.

I did not reorder the engines. #74 says not to and it is right: monolithic last
makes the column look plausible without making it a measurement, and mutation M4
below is that exact shortcut, with the test that catches it.

Three things came along with it.

The sweep's private engine runners are gone in favour of the shared `bench_*`
ones the child dispatches to, so the two engine binaries now measure through one
implementation instead of two that had drifted (the sweep's mapreduce ran without
the `BlankTileStrategy::Emit` and 64-slot buffer config `report`'s uses; on a
gradient raster there are no blank tiles, so the tile counts are unchanged).

The binary grew `--sizes` and `--concurrency`, and `parse_sizes` /
`parse_concurrency` moved out of `report.rs` into the library for both to share.
Nothing could drive this binary over a single cell before, which is most of why a
wrong memory column survived a release: the only way to look at its output was to
run the full grid up to 280 MP.

The bottleneck analysis picks its largest and smallest cells by area rather than
by position, since `--sizes` takes whatever order it is given.

The parent no longer materialises the gradient raster at all, which is a nice
side effect: it used to hold up to 840 MB of source while the engine it was
measuring allocated its own copy.

## What is covered now, and what is not

Covered: `harness::spawn_single_cell` (as before), and the `scalability` binary's
four gradient series, monolithic / streaming / mapreduce / libvips, through the
JSON it publishes.

Not covered, and deliberately: the `pdfium`-gated `streaming-pdf` series still
measures in-process. It is a separate series behind a feature that CI only
`cargo check`s and never runs, `run_single_cell` has no PDF cell to dispatch to,
and `bench_streaming_pdf`'s skip-versus-surface distinction does not survive a
child boundary without inventing an exit-code protocol. Its own printed caveat
now says which series are per-child and that this one is not. It did get quieter
in passing, since the parent it shares a watermark with no longer runs the
engines.

Also not covered, and this one surprised me: nothing in the suite notices if
`spawn_single_cell` stops taking the parent-observed `wait4` RSS (mutation M5,
all green). The child's own `get_peak_rss()` is `RUSAGE_SELF` of a process that
runs exactly one cell, so on an `engines` build the two agree. They diverge only
on the libvips CLI path, where the real `vips dzsave` is a grandchild: `wait4`
folds a reaped descendant's usage in and the child's `RUSAGE_SELF` does not. So
the override is load-bearing for one row on one build and no test says so. I have
left it alone rather than widen this PR; it wants a `vips`-feature test.

## Tests

All in `tests/rss_isolation.rs`, which is where the guard for this defect already
lived. They share one sweep of the real binary (`--sizes 1024x10000,512x360
--concurrency 1`, large first) so the suite pays for it once, about 18 seconds in
a debug build.

`a_sweep_publishes_a_distinct_peak_rss_for_each_engine` is the test that did not
exist. Nothing anywhere asserted that three engines produce three different
figures, which is why a capture was the only thing that could find this.

Its positive control runs first, in the same test: the two rows really are
workloads whose memory differs by an order of magnitude, and the engines' own
`tracked_memory_mb` says so in the same JSON from the same sweep (29.3 MB against
2.34 MB, 12.5x). An assertion that two numbers differ is worth nothing until
something independent says they ought to.

The tall narrow canvas is doing work there. Streaming's footprint is bounded by
canvas width and monolithic's by area, so 1024x10000 buys the order of magnitude
for ten megapixels where the published grid needs about 190 MP to reach the same
ratio.

`a_sweeps_peak_rss_follows_the_canvas_and_does_not_leak_across_sizes` is the
second control and the #157 guard on this path: a column that cannot produce two
different numbers for a 55x change in canvas area cannot be trusted to have
produced two different numbers for two engines either. The small size is swept
after the large one, so a monotonic watermark hands it the large peak.

`a_sweeps_derived_columns_are_recomputed_from_the_published_peak_rss` pins both
derived columns to `peak_rss_mb` for every row.

`the_scalability_binary_answers_the_single_cell_subcommand` drives `--single`
directly, since without it every child would come back empty.

`the_scalability_binary_measures_through_the_child_harness` is the cheap source
tripwire, in the style of `tests/pdf_scalability_series.rs`. It matches on the
call path `libc::getrusage` rather than on `RUSAGE_SELF`, because text cannot
tell code from the prose around it and my first version failed on my own doc
comment explaining the bug...

## Mutations

Every row was run. `cargo test --locked --test rss_isolation` in the pinned
container, mutation applied to the working tree, tree restored and verified clean
after each. Every row rebuilt: `grep -c "Compiling libviprs-bench"` is 1 in all
six logs, so no row is a stale binary reporting the previous verdict.

| # | mutation | rebuilt | distinct-per-engine | canvas / no-leak | derived cols | `--single` | source tripwire | suite |
|---|---|---|---|---|---|---|---|---|
| 0 | none, HEAD as it stands | 1 | green | green | green | green | green | 7 passed |
| M1 | the shipped shape: sweep reads its own `getrusage(RUSAGE_SELF)` and publishes it as every engine's peak | 1 | **RED** | **RED** | green | green | **RED** | 4 passed, 3 failed |
| M2 | cells child-isolated, `peak_rss_mb` per-child, but the two derived columns stand on a shared basis | 1 | green | green | **RED** | green | green | 6 passed, 1 failed |
| M3 | `--single` not wired into `main`, so every child hits the sweep's own parser | 1 | **RED** | **RED** | **RED** | **RED** | **RED** | 2 passed, 5 failed |
| M4 | the forbidden shortcut: shared monotonic watermark plus monolithic measured LAST | 1 | green | **RED** | green | green | green | 6 passed, 1 failed |
| M5 | `spawn_single_cell` drops the parent-observed `wait4` RSS and keeps the child's self-report | 1 | green | green | green | green | green | 7 passed (gap, see above) |

M1 is the defect as captured:

```
assertion `left != right` failed: monolithic and streaming published the SAME
peak RSS (3.76171875 MB) - that is one process-wide watermark reported three
times, not three measurements (issue #74)
  left: 3.76171875
 right: 3.76171875
```

M4 is the row that matters most. Reordering so monolithic runs last makes the
cross-engine test green (streaming 37.75, mapreduce 39.72, monolithic 70.57, all
different, all plausible) while every number is still the same shared watermark.
The cross-size test catches it anyway:

```
peak RSS must follow the canvas: 1024x10000 against 512x360 is a 55x area change
and must move the column by more than rounding; got big=70.31 MB small=70.31 MB
```

M5 is the honest one: no test noticed, and the coverage note above says why.

## Commands

Everything ran in a container on `linux/arm64` (`rust:1.97-bookworm`,
digest-pinned, plus clippy and node), against the core checked out at
`origin/main` 809ee801, into a target directory that started empty. Never the
host toolchain.

| command | exit |
|---|---|
| `cargo build --locked --lib --bins --tests` | 0 |
| `cargo check --locked --lib --bins --tests` | 0 |
| `cargo check --locked --lib --bins --features pdfium` | 0 |
| `cargo check --locked --lib --bins --tests --features polars` | 0 |
| `cargo check --locked --lib --bins --tests --features libvips` | 0 |
| `cargo test --locked --lib --tests` | 0 (36 binaries, 223 tests, 0 failed) |
| `cargo build --locked --release --bin scalability` | 0 |
| `scalability --family engines --sizes 2048x1440` (release, default thread budgets) | 0 |
| `cargo fmt --check` | 0 |
| `cargo clippy --locked --lib --bins --tests` | 0 new warnings in the files I touched |

Clippy under `-D warnings` fails on `build.rs:588` (`collapsible_if`) at this
toolchain, and that is pre-existing on `main`: the same lint fires in
`family.rs`, `harness.rs` and `provenance.rs` too. I did not touch it.

One end-to-end release run, at one size, through the default thread-budget pair
rather than a pinned `--concurrency`:

```
(2.9 MP, c=1) monolithic 21.66  streaming 16.88  mapreduce 19.26   distinct
(2.9 MP, c=8) monolithic 23.82  streaming 16.88  mapreduce 20.81   distinct
```

Against the capture that found this, where all three agreed to seven decimal
places in 20 of 20 groups.

## Not proven

I did not run a real capture. Four lanes are running on this box, so every
number above is a shape check and not a measurement: what I have shown is that
the figures are per-child and that they differ, not what they are. The published
`report/engines/scalability_results.json` still holds the contaminated column and
wants a re-capture on a quiet host.

I did not run the `vips` family end to end (no pinned libvips image here, only a
`--features libvips` compile), so the libvips row's new route through
`Engine::Libvips` is compiled and not exercised.

The `pdfium` cell is `cargo check` only, same as CI.


spdrman:
Three things I found while doing this. None of them got an issue of their own;
they are folded into this PR or written down here so they are not lost.

**The `wait4` override in `spawn_single_cell` is not guarded by anything.**
Mutation M5 removes it, so each row falls back to the child's own
`get_peak_rss()`, and all seven tests stay green, including the two
pre-existing ones. That is not a bug in the mutation: the child's
`RUSAGE_SELF` is a process that runs exactly one cell, so on an `engines`
build the two figures agree. They only diverge on the libvips CLI row, where
the real work happens in a `vips dzsave` grandchild that `wait4` folds in and
the child's `RUSAGE_SELF` does not. So the one line that makes the libvips row
honest has no test. It wants a `vips`-feature test and I have not written one,
because widening this PR into the feature cell I cannot run here would be
trading a proven fix for an unproven one.

**The `streaming-pdf` series is still measured in-process.** It is behind the
`pdfium` feature, which CI only `cargo check`s and never runs, and
`run_single_cell` has no PDF cell to dispatch to. Routing it through a child
needs an exit-code protocol to keep `bench_streaming_pdf`'s
skip-versus-surface distinction, which is more machinery than an unrunnable
series justifies today. The binary's own caveat now names which series are
per-child and says this one is not. It did get quieter in passing: the parent
it shares a watermark with no longer runs the engines or holds the source
raster.

**Clippy's `collapsible_if` fires on `build.rs:588` at Rust 1.97**, which is
what the pinned image carries, and also in `family.rs`, `harness.rs` and
`provenance.rs`. All pre-existing on `main` and all warnings rather than
failures, since the CI clippy step does not pass `-D warnings`. I mention it
only so that nobody reads my `cargo clippy` row as a clean sheet: it is clean
for the files I touched, not for the crate.

And one measurement note. The capture that found #74 is still on disk with the
contaminated column, 20 of 20 groups identical (at 280 MP and 8 threads, all
three engines report `1811.3203125`). This PR does not re-capture: four lanes
are running on this box, so everything here is a shape check, and what I have
shown is that the figures are per-child and that they differ, not what they
are.

PR #81, opened 2026-09-15
K2.2: the engines family emits an archivable document, with repetitions and provenance

### LINKED ISSUE #75: K2.2: the engines family emits a document that can be archived, with repetitions and provenance
The `storage` family produces a document that carries its own provenance, attests its artefacts,
digests four blocks and is refused by its own aggregator when any of that is missing. The `engines`
family produces a bare JSON array of 60 rows with no provenance at all, which is the shape the whole
epic exists to retire.

The page has to show both families. It cannot show one under a provenance callout and the other
under nothing, and it must not silently publish the weaker one.

## What to build

`engines` emits the same document shape `storage` does: `schemaVersion`, `family`, `runner`,
`profile`, `startedAt`/`finishedAt`, `provenance`, `measurement`, `cells`, `invariants`,
`integrity`, and a `runId` derived from the document rather than the clock. Reuse
`storage::document`, `provenance`, `integrity` and `archive` rather than writing a second set. If a
type needs widening to carry a second family, widen it; do not fork it.

The scenario and metric vocabulary differ, so `cells` will carry `wall`, `peak_rss_mb`,
`tracked_memory_mb`, `tiles_per_second` and the two derived columns instead of the storage
scenarios. That is expected. What must not differ is the envelope, the refusal rules, or the digest.

## Repetitions and dispersion

The current sweep measures each (engine, size, concurrency) once. That is the defect the storage
family already fixed and it has to be fixed here too: repetitions, a discarded warm-up, median with
IQR and a bootstrap CI, and the same `confidence` and `lowConfidenceReasons` fields. `harness.rs`
owns the child isolation and the statistics already.

## Invariants

`tiles_produced` is exact and reproduces; make it an invariant with an equality verdict rather than
a timing. Check whether `output_bytes` is available here and whether it reproduces across
repetitions before declaring it one. The storage capture found that `allocated_bytes` is **not**
invariant (it differs between repetitions and generation was refused on two cells), so do not assume
a field is exact because it looks like it should be.

## Coordinate with #74

#74 fixes the peak RSS contamination by routing `scalability.rs` through
`harness::spawn_single_cell`. That lane lands first and you build on it. Do not duplicate its fix,
and do not publish a `peak_rss_mb` that came from a process-wide watermark.

## Acceptance

- `storage-aggregate --check` and `--archive` accept an `engines` document, with the same rules.
- A document missing provenance, built in debug, or taken under emulation is refused, and a test
  proves each refusal fires on an `engines` document specifically rather than only a storage one.
- Every cell carries `reps`, `samples`, a median of them, and a `confidence`.
- `tiles_produced` is an invariant with an equality verdict.

spdrman:
The `storage` family emits a document that carries its own provenance, attests its artefacts,
digests four blocks and is refused by its own aggregator when any of that is missing. The `engines`
family emitted a bare JSON array of sixty rows with no provenance at all. The page has to show both
and cannot put one under a provenance callout and the other under nothing.

So `engines` now emits the same envelope, through the same code. Not a second document type: this
crate already carries one port of causl's chart code that diverged, and a forked `Document` would
put the same divergence inside the crate.

## What changed shape rather than being copied

`Measurement.reps` is a map keyed by scenario kind, because `storage` declares `{generate, read}`
and `engines` declares `{pyramid}`. A `BTreeMap`, so the storage block still serialises
`{"generate": N, "read": M}` byte for byte and the rule that compares it against `resolved.reps` is
untouched. `Measurement::probed_from` takes the half a family decides and fills the clock probe, the
bootstrap, the tie band and the confidence threshold from the shared constants, so two families
cannot publish intervals computed different ways under one set of field names.

`CellReport` takes four labels instead of storage's own `Backend` and `Cell`. The vocabularies
differ, the shape does not: which producer, at which scale, over which source, in which cell.

`storageAttested` is now `attested`. It was a field name that had turned into a family name; the
rule and the refusal are identical for both.

`Provenance::to_storage_block` and its two siblings lose the `storage` in their names. Nothing in
them ever knew which sweep was asking.

The archive gets a directory per family. Both families derive a run id from `startedAt`, the
measured commit and an environment bucket, and none of those is the family, so a `storage` and an
`engines` sweep started in the same second against the same commit derive the **same id** and in
one directory the second reads as a collision with the first. Putting the family into the id is the
other fix and it is the wrong one: the id is what a page keys an era on, and a family is not part
of an era.

`RunMetrics` grows `artefact`: what the run's pyramid was observed to be on disk, walked with the
same `storage::occupancy` the directory backend uses. Untimed, taken after `wall_time` and before
the sink directory is reclaimed. `None` on every libvips path, which writes through `dzsave` into a
directory this crate does not own.

## Repetitions, and the peak RSS

Every `(engine, cell)` takes a discarded warm-up and then N timed repetitions, each in its own child
through `harness::spawn_single_cell`, with the engine order rotated between repetitions so drift
over a cell's window hits the three equally.

That is also what makes the memory column honest. The sweep behind the first full capture ran all
three engines in one process against `getrusage(RUSAGE_SELF)`, a monotonic process-wide watermark,
which is why that capture reports byte-identical peak RSS for all three engines in twenty of twenty
groups. Measured here, at `4096x2880` with four fresh children each:

| engine | peak RSS (MB) |
|---|---|
| monolithic | 78.85, 78.72, 78.79, 78.76 |
| streaming | 50.75, 50.41, 50.78, 50.86 |
| mapreduce | 56.83, 56.34, 56.94, 56.27 |

Three engines, three memory models, three different numbers.

## The invariants

`tiles_produced` is an invariant with an equality verdict, and so are `output_bytes`,
`filesystem_entries` and `directories`, all walked off the real sink directory. A cell whose
repetitions disagree about one of them is refused with the field named rather than averaged: inside
one commit that is a defect, not a delta.

`allocated_bytes` is measured and not published, and I checked rather than assumed. It did not
wobble: 24 fresh children at `1024x720` and 12 at `4096x2880`, three engines each, one value every
time. The reason to leave it out is what the number is. It is `st_blocks * 512`, so it answers to
the filesystem's allocator rather than to the engine, and the storage family's first capture caught
it moving between two repetitions of one generation. A claim a block size can falsify is not a claim
about the engine.

## Attestation

An engine may carry `attested: true` only when it wrote a pyramid, wrote the same one every
repetition, and agrees with at least one sibling about the per-level tile grid and the tile count.
The evidence is a walk of the real sink directory done in the child that ran the engine. One engine
alone is never attested: it has nothing to agree with, and the whole reason the grid is worth
checking is that three engines walking one plan must produce the same tiles.

## Composed with K2.1

`fix/74-engines-rss` is not merged yet, so this branch stays based on `main` rather than pinning an
unmerged head. It is composed and checked all the same, on a scratch branch: the merge is clean
(only `src/lib.rs` overlaps, in two different regions) and the composed tree builds and passes the
whole suite.

The two lanes meet at `harness::spawn_single_cell`, which is the route this document's
`peak_rss_mb` already takes, so there is nothing here to rework around their fix. The one thing to
carry forward is that their change removed a parent-side 840 MB gradient raster, so **any memory
figure measured before that lane is not comparable to one measured after it**. The numbers in this
PR are from `--single` children, which never held the parent's copy, but the caveat matters for the
archive: an `engines` run archived before K2.1 and one archived after are not the same experiment on
the memory columns.

## The noise floor

Seven repetitions of a cell measure the noise inside one wall-clock window and
nothing else. What they cannot see is drift *across* the sweep, which is thermal,
a neighbour waking up and the page cache filling, and that is the drift the
epic's own headline number came out of: a free replicate pair in the committed
exports where p99 moved 74.5% on an idle host running identical code.

The storage family opens and closes on one control cell for exactly that reason.
`engines` now does too, on `full` and `xl`, and publishes the same `replicate`
block. `replicate::schedule` is generic over the cell type and `block_for_cell`
finds the control's two visits by the cell's own spec instead of by
`(scale, source)`. That second half is not cosmetic here: two engines cells can
share a tile count and a source and differ only in their thread budget, and a
filter that could not tell those apart would compute a spread between two
different measurements and publish it as drift.

## Four things found on the way, folded in

**`--allow-dirty` could never work, in either family.** The aggregator refuses a document whose
`provenance.allowDirty` is true and whose cells do not each carry `dirty: true`, and nothing ever
wrote that field. I found it by setting the variable: *"provenance.allowDirty is true but 18 of 18
cells do not carry dirty: true"*. Both sweeps now stamp it from their provenance, so the flag
satisfies the rule it exists for. Stamped from the provenance and not from the flag, so a clean tree
does not acquire a caveat it has not earned.

**The `runId` was always `null`.** Both families left it for the aggregator to derive at archive
time, so a reader who never ran the aggregator could not name the run. Both now stamp it through the
same `archive::run_id` the aggregator files them under.

**One archive directory for two families**, as above.

**Two tests in `storage_registry` shared one scratch directory.** It was keyed on
the test binary's pid, which libtest shares across the threads it runs the tests
on: one test finishes, removes the directory, and the other's `storage` child is
still writing into it. The child exits 1 and the failure reads as *"the storage
binary exited exit status: 1"* with nothing to say why. It only shows up under
load, which is where I found it, twice, in this lane's own gate runs. A directory
per call fixes it.

## The mutation table

33 rows, every one run, in a comment below. 33 RED, three surviving on purpose (the two `run_sweep`
wiring lines and one of them twice), one INCONCLUSIVE that is an artefact of the driver rather than
a gap.

It found three real holes in my own tests and all three are fixed here: a facet-key assertion whose
fixture sat exactly on the mutation's fixed point, a hole test that checked one of the two derived
columns, and a replicate fixture that could not distinguish the mutation it was written for.

## What the lanes downstream need to know

Four names changed or arrived, and #76 and #77 read all of them:

- `family` is `libviprs-engines`, `runner` the same. `storage`'s are unchanged.
- a cell's attestation key is `attested`, not `storageAttested`, in **both** families.
- `measurement.reps` is an object keyed by scenario kind. `storage` still writes
  `{"generate": N, "read": M}`; `engines` writes `{"pyramid": N}`.
- an archived run lives under `archive/<family>/`, so `archive/storage/` and `archive/engines/`,
  each with its own `index.json`.
- `runId` is filled by the producer now, not only by the aggregator.

## What I did not do

`modelled` stays `[]`. It is the storage family's two declared cost models and there is no engines
analogue; a remote store's charge is not a question about which engine wrote the tiles.

I have not run a `full` or `xl` sweep. The `ci` profile runs in the test suite and in the container,
and the arithmetic above comes from `--single` children driven directly, but how long 384 children
take and how much memory the two `xl` canvases want in a child are **not proven**.

Nothing ran on x86_64. This is arm64 only.

`run-bench.sh` is not wired to the new binary. The `engines` binary is in the README and in the
Docker `engines` stage, which is what K2.4 and K2.5 need; adding a command word to `run-bench.sh`
would collide with #74, which is editing `scalability.rs` in the same wave.

Closes #75


spdrman:
## Mutation table

Every row is a one-place change to the **product** (never to a test), the tests that had to go red
against it, and whether cargo actually rebuilt: a row with no `Compiling libviprs-bench` in its
output is INCONCLUSIVE, because cargo may have served a cached binary and the result would say
nothing about the mutation. Run in a container, arm64, one row at a time, each reverted from a byte
snapshot taken before the edit.

| # | mutation | file | test that had to go red | rebuilt | verdict |
|---|---|---|---|---|---|
| M1 | half-done rename: CELL_FIELDS still says storageAttested | `src/storage/document.rs` | `the_engines_document_is_the_same_envelope_the_storage_one_is` | yes | **RED** |
| M2 | drop the resource-cost column | `src/engines/mod.rs` | `a_cell_publishes_all_six_columns` | yes | **RED** |
| M3 | single shot: publish only the first repetition | `src/engines/mod.rs` | `every_cell_carries_reps_samples_a_median_of_them_and_a_confidence` | yes | **RED** |
| M4 | throw the warm-up away without recording it | `src/engines/mod.rs` | `the_discarded_warmup_is_published_and_is_not_in_the_samples` | yes | **RED** |
| M5 | derive the efficiency column from a summary rather than per repetition | `src/engines/mod.rs` | `the_derived_columns_are_derived_once_per_repetition` | yes | **RED** |
| M6 | the old to_point: write 0.0 where there is no denominator | `src/engines/mod.rs` | `a_derived_column_with_no_denominator_publishes_no_sample_rather_than_a_zero` | yes | **RED** |
| M6b | the old to_point on the efficiency column specifically | `src/engines/mod.rs` | `a_derived_column_with_no_denominator_publishes_no_sample_rather_than_a_zero` | yes | **RED** |
| M7 | average a moved invariant instead of refusing the cell | `src/engines/mod.rs` | `tiles_produced_is_an_invariant_with_an_equality_verdict` | yes | **RED** |
| M8 | publish the filesystem allocator's answer as an invariant | `src/engines/mod.rs` | `allocated_bytes_is_measured_and_not_published_as_an_invariant` | yes | **RED** |
| M9 | a run id off the clock | `src/storage/document.rs` | `the_run_id_is_derived_from_the_document_and_never_from_the_clock` | yes | **RED** |
| M10 | publish the smoke profile | `src/engines/cells.rs` | `only_a_repeated_profile_is_publishable` | yes | **RED** |
| M12 | the producer as it stood: never stamp the dirt onto a cell | `src/storage/document.rs` | `a_dirty_tree_stamps_the_caveat_onto_every_cell` | yes | **RED** |
| M13 | stamp attested: true and move on | `src/engines/attest.rs` | `the_verdict_moves_with_what_was_observed` | yes | **RED** |
| M14 | let one engine attest itself | `src/engines/attest.rs` | `one_engine_alone_is_never_attested` | yes | **RED** |
| M15 | compare the grid and not the tile count | `src/engines/attest.rs` | `the_grid_and_the_tile_count_are_both_compared` | yes | **RED** |
| M16 | compare the tile count and not the grid | `src/engines/attest.rs` | `the_grid_and_the_tile_count_are_both_compared` | yes | **RED** |
| M17 | an empty directory counts as an observation | `src/engines/attest.rs` | `an_empty_artefact_is_unobserved_rather_than_small` | yes | **RED** |
| M18 | the sweep leaves its own provenance empty | `src/engines/mod.rs` | `the_sweep_writes_a_document_its_own_aggregator_accepts` | yes | **RED** |
| M19 | the aggregator still reads the storage family's field name | `src/storage/archive.rs` | `the_aggregator_binary_checks_and_archives_an_engines_document` | yes | **RED** |
| M20 | one archive directory for both families | `src/storage/archive.rs` | `an_engines_document_is_filed_apart_from_a_storage_one` | yes | **RED** |
| M20b | the binary's default root ignores the document's family | `src/storage_aggregate_main.rs` | `the_aggregator_files_an_engines_document_under_its_own_family_by_default` | yes | **RED** |
| M21 | a document that does not say whether it was translated is admitted | `src/storage/archive.rs` | `an_engines_document_without_provenance_is_refused` | yes | **RED** |
| M21 | a document that does not say whether it was translated is admitted | `src/storage/archive.rs` | `an_engines_document_taken_under_emulation_is_refused` | NO | **INCONCLUSIVE** |
| M22 | a document that does not say whether it measured a release build is admitted | `src/storage/archive.rs` | `an_engines_document_built_in_debug_is_refused` | yes | **RED** |
| M23 | stop checking attestation on ok cells | `src/storage/archive.rs` | `an_engines_document_with_an_unattested_ok_cell_is_refused` | yes | **RED** |
| M24 | let the invocation and the measurement block disagree about repetitions | `src/storage/archive.rs` | `an_engines_document_whose_reps_disagree_is_refused` | yes | **RED** |
| M25 | admit a document whose scenarios still say "all" | `src/storage/archive.rs` | `an_engines_document_with_unresolved_scenarios_is_refused` | yes | **RED** |
| M26 | a cells digest over a key no document has, so it can never move | `src/storage/integrity.rs` | `the_four_digests_over_an_engines_document_move_when_its_numbers_do` | yes | **RED** |
| M27 | overwrite an archived run when a second one derives the same id | `src/storage/archive.rs` | `an_engines_run_that_collides_is_refused_rather_than_overwritten` | yes | **RED** |
| M28 | a sweep with no control at either end | `src/engines/cells.rs` | `a_publishable_profile_opens_and_closes_on_its_control_cell` | yes | **RED** |
| M30 | stamp the attestation as a constant in run_sweep (expected to survive) | `src/engines/mod.rs` | `the_sweep_writes_a_document_its_own_aggregator_accepts` | yes | **SURVIVED** |
| M30 | stamp the attestation as a constant in run_sweep (expected to survive) | `src/engines/mod.rs` | `tiles_produced_is_an_invariant_with_an_equality_verdict` | yes | **SURVIVED** |
| M31 | drop the noise floor from the document (expected to survive) | `src/engines/mod.rs` | `the_sweep_writes_a_document_its_own_aggregator_accepts` | yes | **SURVIVED** |
| M32 | publish the engine's own accounting under the operating system figure's name | `src/engines/mod.rs` | `the_two_memory_columns_are_two_different_bases` | yes | **RED** |
| M33 | read the parent's process-wide watermark instead of the child's | `src/engines/mod.rs` | `a_small_cell_measured_after_a_large_one_reports_the_small_figure` | yes | **RED** |
| M29 | find the control's two visits without checking which cell they are | `src/storage/scenarios/replicate.rs` | `the_replicate_block_is_the_spread_between_the_two_ends_of_the_sweep` | yes | **RED** |

totals: {'RED': 32, 'INCONCLUSIVE': 1, 'SURVIVED': 3}

### The two that survived and were fixed

**M11, the facet key, survived a fixed point.** The mutation returns a hard-coded `25` from
`planned_tiles`, and the test compared one cell's key against its own plan. `1024x720` really does
plan 25 tiles, so the fixture sat exactly on the mutation's fixed point and the assertion could not
fail. It now walks three canvases and requires three distinct keys.

**M6, the hole in a derived column, survived a half-covered assertion.** The test checked
`tiles_per_second_per_mb` and not `resource_cost`, so publishing `0.0` where there is no denominator
got through on the second column. That is the worse of the two: lower is better on resource cost, so
a hole written as a zero reads as the *best* possible score.

**M29, the replicate control, survived an indistinguishable fixture.** With the control closing the
sweep, "the last row for this metric key" and "the control's last row" are the same row, so dropping
the cell check changed nothing. The fixture now also runs the case where they differ.

All three are RED after the fix, and both fixes are in this branch.

### The three that survive on purpose

`M30` and `M31` mutate the two lines in `run_sweep` that copy the attestation verdict onto a row and
hang the noise floor on the document. In the `ci` profile every cell really is attested and there is
no control cell, so a constant there is indistinguishable from a correct run and no assertion over
that document can tell them apart. Catching them needs a `full` sweep in a test, which is minutes
rather than seconds. They are published as surviving rather than quietly dropped.

`M21`'s second entry is an artefact of how the table is run rather than a real gap: the row drives
two tests, the first invocation carries the rebuild witness and goes red, and the second reuses the
binary the first one just built. The test failed; the witness for that particular invocation is
absent, so the driver reports it honestly as INCONCLUSIVE.

PR #87, opened 2026-09-15
A row is a replicate when it is the declared cell, not when it resembles it

### LINKED ISSUE #85: Replicate detection matches on tile count and source, so the whole eight-thread arm is filed as replicates
The importer files a row as a replicate of the declared replicate cell when it matches on tile count and
source. In the `engines` family every `+c8` cell shares both with its `+c1` twin, so the whole eight-thread
arm gets swallowed.

Measured on the real archived engines run `20260915T021911Z-809ee801...-a7bac4a9`: **162 rows arrive as
replicates, 18 of them genuine and 144 a separate experiment.** The run declares `512x360@256+c1` as its
replicate cell; `512x360@256+c8` is a different measurement of a different thing.

They differ measurably, so they are not duplicates that happen to be filed oddly. At the largest cell
mapreduce reads 348.96 MB at one thread and 365.16 MB at eight.

## This is a key collision, and it is the second time

K2.3 found the same root cause from a different direction and wrote it down: tile count is not unique across
cells, because `21851` is both `8192x8192@64+gradient` and `+noise`. Cell identity needs enough of the label
to actually identify a cell. Replicate detection is the second consumer to be bitten by it.

## What it costs

The eight-thread arm is the concurrency story. Filed as replicates it stops being renderable at all, and the
replicate spread it contaminates is the number the page uses to decide whether a delta is noise. So one
collision quietly damages both the thing being measured and the instrument measuring it.

The page recovers by reading rows that name a cell other than the declared one as the cells they say they are,
and it says how many and why. That is a consumer working around a producer, and it belongs here instead.

## What to change

Key replicate detection on the full cell identity the row already carries rather than on a tile count plus a
source. The declared replicate cell names itself; a row is a replicate when it *is* that cell, not when it
resembles it.

## Acceptance

- The engines run yields 18 replicate rows, not 162.
- A test builds two cells differing only in concurrency and asserts they are not treated as replicates of one
  another. It goes red today.
- A test asserts the same for two cells differing only in source, which is K2.3's case and is the same bug.

spdrman:
Closes #85

The importer filed a row as a replicate of the declared replicate cell when it
matched an already-published row on the section key and the tile count. Neither
of those names a cell. A section is how the page groups rows for drawing, and a
tile count is a magnitude, so two genuinely different cells that happen to share
both were read as one cell measured twice.

In the `engines` family that is every `+c8` cell against its `+c1` twin, and the
arm it swallowed is the concurrency story. On the real archived run
`20260915T021911Z-809ee801...-a7bac4a9`: **162 rows arrived as replicates, 18
genuine and 144 a separate experiment.** They are not duplicates filed oddly, they
differ measurably: at the largest cell mapreduce reads 348.96 MB at one thread and
365.16 MB at eight. And it cost twice over, because the replicate spread is the
number the page compares a delta against to decide whether it is noise, so one
collision damaged both the thing measured and the instrument measuring it.

This is the second consumer bitten by one root cause. K2.3 found it from the
other direction and put `source` into the section key because `21851` is both
`8192x8192@64+gradient` and `+noise`. Cell identity needs enough of the label to
identify a cell.

## What changed

Identity is now the producer's own fields, named in the config as
`producer.rowIdentityFrom`: `backend`, `cell`, `source`, `key`. The run declares
the cell it measured twice and that cell names itself, so a row is a replicate
when it **is** that cell rather than when it resembles it.

Four choices in there, each with a test behind it rather than only a comment:

* the raw `backend` and not the renamed series, because a rename is a display
  decision and identity must not move when a label does;
* no scale, because the cell name already determines it. A producer emitting one
  cell name at two tile counts is filing a defect, and the loud version of that
  is two rows colliding rather than two rows quietly splitting;
* a JSON array rather than a joined string, so `a` and `b · c` cannot spell what
  `a · b` and `c` spells;
* every field in the list is now required on every measured cell. An identity
  with a hole in it makes every row look like the same row, which reads as a
  large confident replicate count rather than as a failure.

`archive/engines/` now holds the run, because the suite runs against real
archived documents and nothing else, and the directory had a README describing a
shape with no document in it.

## The page can drop its workaround

`benchmarks/tools/render-latest.mjs` recovers rows filed as replicates of a cell
they are not, and says how many and why. After this, `misfiled` is empty on the
engines run and `counts.recovered` is 0, so the recovery path and the note above
the chart are both inert. That was a consumer compensating for a producer and it
belongs here, so the page lane can take it out.

## Mutation table

Every row was run. The witness is `sha256sum` over the three files node reads,
taken **inside the container** immediately before and after the mutation, so a
row cannot be a colour reported against bytes that never moved. The suites are
`node --test tools/publish/*.test.mjs` (68 tests) and `tools/contract` `npm test`
(34), both on `node:22-bookworm-slim`, `--platform linux/arm64`.

| # | mutation | witness (import-run · publish cfg · contract cfg) | result |
|---|---|---|---|
| M1 | `rowIdentity(cell)` back to `` `${library}/${scenarioOf(cell)}@${cell[SCALE_FROM]}` ``, the defect restored | `79f8199 → e2cf35a` · `cd0aba3` · `162edd5` | **RED**, 4 of 68: the engines count, the concurrency pair, the section-key decoupling, the renamed-series pair |
| M2 | `cell` dropped from `producer.rowIdentityFrom`, both configs | `79f8199` · `cd0aba3 → 3f1d7ca` · `162edd5 → c6551bc` | **RED**, 9 of 68, four of them pre-existing storage tests |
| M3 | `source` dropped from `producer.rowIdentityFrom`, both configs | `79f8199` · `cd0aba3 → 4153f72` · `162edd5 → 10642de` | **RED**, 1 of 68: the source test, and only the source test |
| M4 | `...ROW_IDENTITY_FROM` removed from `REQUIRED_CELL_FIELDS` | `79f8199 → 9c681bc` · `cd0aba3` · `162edd5` | **RED**, 1 of 68: the missing-cell refusal |
| M5 | identity joined with ` · ` and absent fields filtered out, instead of encoded | `79f8199 → d59240e` · `cd0aba3` · `162edd5` | **RED**, 1 of 68: the separator pair. **Survived the 66-test suite it was first run against** |
| M6 | `rowIdentityFrom` deleted from `tools/contract/config.json` | `79f8199` · `cd0aba3` · `162edd5 → df7f420` | **RED**, 1 of 68 and 1 of 34: the key is anchored in the config, not only in a default |
| M7 | identity reads `seriesId(cell)` where it reads `backend` | `79f8199 → e1fa1b4` · `cd0aba3` · `162edd5` | **RED**, 1 of 68: the renamed-series pair |

M3 is the one worth reading twice. It is the only mutation the source test
catches, and the source test is **green before this change** for a reason I would
rather write down than be reassured by: it passes today because
`sections.scenarioFrom` happens to list `source`, which is a decision about
drawing. The test that proves the identity no longer borrows that decision is the
one that runs the storage import with `scenarioFrom: ['key']`, the importer's own
default, and that one is red before the change at 142 replicates against 48.

**M5 survived the first time it was run**, against the suite as it stood: joining
the identity fields into one string instead of encoding them passed all 66 tests.
So did reading the renamed series instead of the backend. Both are choices I had
argued for in a comment with nothing holding them up, which is how a comment ends
up being the only thing a decision rests on. Two tests were added for them and
the mutation is red now.

## Folded in

No new issues, per the lane rules. Everything found along the way is in this PR:

* **the field-name defaults were written out twice**, once for the shape gate and
  once for the reader in section 7, so the gate could end up checking for a field
  nothing reads while the reader fell back to a field nothing checked. Read once
  now.
* **the contract suite now checks `rowIdentityFrom` against the producer's real
  cell fields.** Naming a field no cell carries neither refuses nor throws, it
  collapses every row into one, which is the same failure mode as the defect.
* **the workflow's positive control now covers both archived runs.** It asserts
  the storage document is in the repository so that a suite full of refusals
  cannot pass by having nothing to import; the engines document needs the same.
* **a test of mine was reading a triple as a pair.** 21851 is a tile count and
  three `generate.wall` rows across two backends carry it, so the assertion named
  a backend. Same point as the test it lives in, arriving from a third direction.

## Not changed

No Rust. The producer is right about what it measured and spells its cells
correctly in both families; this was entirely a reader keying on the wrong thing.


spdrman:
Four things turned up while I was in here and all four are in this PR rather than
in new issues.

1. **The field-name defaults were written out twice.** `sections.scenarioFrom ??
   ['key']` and friends appeared once in the shape gate and again in the reader,
   so the gate could end up requiring a field the reader no longer looks at, or
   the reader falling back to one the gate never checked. They are read once now,
   above the gate that needs them first.
2. **Nothing checked `rowIdentityFrom` against the producer's real cell fields.**
   The contract suite already checks `sections.*` and `samples.carry` that way,
   and identity is the field list where naming a non-existent field hurts most:
   it does not refuse and it does not throw, every row simply becomes the same
   row. Added to `config.json names only things the producer actually emits`.
3. **The workflow's positive control covered one archived run.** It exists so a
   suite full of refusals cannot pass by having nothing to import, which applies
   to `archive/engines/` exactly as it does to `archive/storage/`.
4. **One of my own tests was reading a triple as a pair.** 21851 is a tile count,
   and three `generate.wall` rows across two backends carry it, so the assertion
   picked up gradient, noise and a second noise. It names a backend now. The test
   is about tile counts not identifying cells, so it was making its own point
   against itself.

And one handover. The page recovers rows filed as replicates of a cell they are
not (`benchmarks/tools/render-latest.mjs`, the `misfiled` / `counts.recovered`
path and the note it renders above the chart). After this lands, `misfiled` is
empty on the engines run and `recovered` is 0, so both are inert and the page
lane can take them out. The 288 rows the page gets are the same 288 it
reconstructs today, so nothing downstream changes shape: they arrive filed
correctly instead of being recovered.

PR #101, opened 2026-09-15
Gate the publish on the load the sweep did not put there

### LINKED ISSUE #100: The quiet gate counts the sweep's own load, so the busiest family is the one it calls noisy
`refuse.majorityNoisyCells` in `tools/publish/import-run.mjs` refuses a run when more than half its measured cells carry `machineLoad.quiet: false`, and `quiet` is derived from `loadAvg1m / cores` at the moment the cell was measured. The load average does not distinguish the benchmark's own threads from anything else on the box, so the gate is partly measuring the thing it is trying to protect.

The 2026-09-15 NAS capture shows the shape clearly. Same machine, same six cores, the two families back to back:

| family | cells | noisy | median load |
|---|---|---|---|
| storage | 539 | 0 (0.0%) | 4.51 |
| engines | 558 | 108 (19.4%) | 5.13 |

Storage is largely single-threaded I/O work. Engines runs monolithic, streaming and mapreduce, and the thread-scaling cells deliberately saturate the box. So the family that reads noisy is the one whose job is to put load on the CPU, and the noisy cells' median load average is 6.24 against six cores, which is roughly one runnable thread per core: a benchmark working, not a benchmark being interfered with.

Neither run was refused this time, so nothing is broken right now. What bothers me is the direction of the error. Turn the engines sweep up (more threads, more cells, a busier profile) and it walks toward the refusal on its own, while an actually contended host running a quiet family sails through. The gate gets stricter as the measurement gets more demanding, which is backwards.

I am raising this rather than fixing it, because the replacement needs a decision rather than a patch.

What I would pair, for discussion:

1. **A pre-run baseline.** Read the load average before the sweep starts, on the settled machine, and record it on the envelope. That number is other people's work by construction, because none of ours has started. The existing settle loop already reads it, so this is mostly plumbing.
2. **The replicate dispersion that is already published.** Every cell now carries `spreadPct`, `driftPct` and `residualPct` across `measured_cells + 1` placements with the `t(n-1)·√(1+1/n)` estimator. Interference shows up there as dispersion the estimator cannot explain, and unlike a load average it is evidence from the measurement rather than from the host.

A run would then be refused when the machine was already busy before we touched it, or when the replicates disagree more than the estimator allows, and not merely because our own benchmark used the CPU it was given.

Worth keeping either way: the current rule's refusal message is good, and whatever replaces it should still say how many cells and what the load was.

Acceptance:

- A run whose only load is its own sweep is not refused, proved on a capture whose cells report a load average at or above one per core.
- A run measured on a host that was busy before the sweep started is refused, proved by a fixture rather than by waiting for one to happen.
- `refusal-enumeration.test.mjs` still enumerates every refusal, including whatever this becomes.

spdrman:
Closes #100

The quiet gate counted cells recording `machineLoad.quiet: false` and refused a run when more than half of them did. That was measuring the sweep as much as the machine. A one-minute load average has a one-minute memory and a sweep spends every second of it working, so by the second cell the number is mostly our own recent work, and the busier the family the louder it reads.

The 2026-09-15 capture said it plainly. Back to back on one six-core box:

| family | cells | noisy | median load | median noisy load |
|---|---|---|---|---|
| storage | 539 | 0 (0.0%) | 4.51 | n/a |
| engines | 558 | 108 (19.4%) | 5.13 | 6.24 |

Neither run was refused, so nothing was broken. What bothered me is the direction of the error: turn the engines sweep up and it walks into its own refusal, while a genuinely contended host running a quiet family sails through.

I also checked the thing I assumed, and I had it wrong. I expected the noise to be the `+c6` cells alone, and it is not: 54 of 144 `+c6` cells and 54 of 414 `+c1` cells read loud, with `+c1` reaching 6.29 on six cores while running one worker. So the per-cell load is not our concurrency, it is the sweep's own recent history decaying through the average, and no arithmetic on that number will separate our work from anyone else's.

## What replaces it

The runner samples the machine as its first act, before it has measured anything or spawned a child, and carries that on the document as `startingLoad`. At that instant none of the load is ours, which makes it the one reading in the document that is other people's work by construction. The line on it is the same line `MachineLoad` has always drawn, one runnable thread per core: no number was invented, it moved to a place where it means what it says.

Three states, kept apart on purpose:

* **key absent** is a run from before this change, and it keeps the majority rule it was published under. The site re-verifies its whole published history on every ingest, so judging those runs by a field they could not carry would refuse runs already on the page.
* **key present and null** is a runner that skipped its own first act, and it is refused. serde writes `None` out rather than dropping the key, so folding the two absences together would make deleting one line of the producer a way to turn the gate off.
* **numbers that will not read** is an empty reading, and refused as one. There is also a cross-check that `quiet` agrees with the `contentionPerCore` beside it, because a verdict that does not follow from its own numbers came from somewhere else.

## What I did not do

The issue proposed pairing the baseline with the replicate dispersion the run already publishes. I looked and did not. On a perfectly good engines run `spreadPct` sits at 40 to 46% on wall and `residualPct` at 43 to 48%, so any threshold I picked would be a number I made up, which is what the importer's own comment argues against.

The per-cell loads stay in the document and stay a low-confidence reason on the cell. They are honest about what they are, a load average while this cell ran, and a reader should have them. They are simply not a thing to refuse a publish over. Changing that annotation would move the page and belongs to its own issue.

## Proof

Every new test was mutation-verified rather than assumed. Five mutations of the importer, each caught by the test that names it:

| mutation | caught by |
|---|---|
| busy refusal deleted | a run measured on a machine that was already busy is refused |
| unreadable check disabled | a starting load that could not be read is refused |
| agreement check disabled | a starting load whose verdict does not follow from its numbers |
| `?? null` instead of `Object.hasOwn` | a runner that left it null is refused, not read as absent |
| old rule kept unconditionally | a run that loaded the machine itself, from a quiet start, is published |

And two on the producer: a constructor that samples for itself fails `a_document_that_never_ran_claims_no_starting_load`, and deleting the assignment in `run_sweep` fails `the_sweep_records_the_machine_before_it_measures_anything`.

What the tests cannot see is a sample taken at the END of a sweep. That is held by the call site instead: the sample is taken before `Document::new_for` runs, so it necessarily precedes every cell the document holds. I would rather say that than pretend an assertion covers it.

No schema bump. Nothing existing changed meaning, and bumping would split the page's x-axis on an era break that did not happen.

## Local gate

`cargo fmt --check`, the four `cargo check` feature cells and `cargo clippy --features libvips` all green (that is this repo's pre-commit hook, run rather than skipped). `cargo test --lib --tests --no-fail-fast`: 41 binaries ok, 2 failed, 6 tests, all of them the macOS emulation-probe family (`REFUSED[emulated]: the probe answered "unknown"`). I stashed the change and got the identical failure set on `main`, so they are this machine rather than this branch, and CI's native Linux runner is what can answer them.

Also green: `tools/publish` 89/89, `tools/contract` gate (sync, parameterize, 35 + 23 tests), `capture-nas.test.mjs` 15/15, `test_capture.py` 27/27.

## Two passengers

`capture.py` and `capture-nas.sh` both described the gate they were settling for, and both sentences were stale. The settle threshold and the gate line are deliberately different, half the cores against one per core, so a settled machine has headroom instead of sitting on the line, and the comments now say that.

And `cpu_model`'s macOS branch failed `clippy::collapsible-if` on rust 1.98, which blocked this repo's own pre-commit hook on any Mac. It sits inside `#[cfg(target_os = "macos")]`, so CI never compiled it and never saw it. A gate people have to skip is a gate that dies, so I collapsed it.

libviprs-org carries its own copy of these rules and needs the same split, which is libviprs/libviprs-org#83.


spdrman:
One thing the new rule does not catch, said out loud rather than left for somebody to discover.

The old rule would refuse a run that started on a quiet machine and then had somebody's build land on it halfway through, because the second half of the cells would read loud. The new rule reads one sample, at the start, so it would publish that run.

I went looking for a way to keep both and did not find one I could defend. The obvious shape is "refuse a cell whose load exceeds the starting load plus what this sweep can produce", and the second term is the problem. A `+c6` cell on six cores legitimately adds six, on top of a starting load of two, on top of a one-minute average that is still carrying the previous cells. Every candidate line I tried either passes a genuinely contended run or refuses the engines family doing its job, and choosing between them means picking a number out of the air, which is what this whole issue is about.

What would actually settle it is a measurement rather than an arithmetic guess: sample the load again at the END of the sweep and compare the two, or sample it from outside the measuring process. Both are real work and neither belongs in this PR.

So the honest summary of the change is narrower than "the gate got better": it stopped refusing runs for their own load, and it did not gain a way to notice work that arrives mid-sweep. The run still records every per-cell load and still marks those cells low confidence, so the evidence is on the page for a reader who wants it.

spdrman:
Second commit, and it came from running the thing rather than reasoning about it.

I drove the real `engines` binary through the new gate and it published this:

```
startingLoad: {"cores": 10, "loadAvg1m": null, "contentionPerCore": null, "quiet": null}
```

My own rule refuses that as "the machine was looked at and could not be read", and the refusal is false. The machine will say.

`MachineLoad::sample` read `/proc/loadavg` itself, so it was Linux only, while `provenance::load_average` in the same crate already handles both platforms and goes through `getloadavg` on macOS. Two readers of one quantity, disagreeing on one document: every cell published a null load while the provenance block beside them carried a real reading. Under the old rule it never surfaced, because cells recording `quiet: null` were skipped by the majority count. The new rule turns that silence into a refusal, so the duplication stopped being cosmetic the moment the gate started reading it.

`MachineLoad::sample` now goes through the one reader. The same binary publishes `4.92 on 10 cores, contention 0.492, quiet: true`, and the importer's verdict on that document carries none of the load rules, only the ci-profile, debug-build, dirty-tree, unarchived and emulation-unknown ones this machine earns.

The producer test lost its `#[cfg]` split and is better for it: it now asserts the load, the contention arithmetic and the `quiet` agreement on every platform, rather than asserting on Linux and shrugging elsewhere. That shape is what would have hidden this in the first place, which is the third time in this epic that a skip wearing a pass's colour has cost something.

Local gate after the second commit: fmt, four check cells and clippy green, `tools/publish` 89/89, 41 test binaries ok and the same 2 failing on the macOS emulation probe that `main` fails on.

